PR #6, opened 2024-08-28
Updates refreshTokenHalder to use the corectly the context instead of HTTP header

emmdim:
… headers


p4u (approved):
Good!
Can you check that the rest of the methods use correctly the context to verify the user please?

emmdim:
@p4u I confirm that all the `auth`, `user` and `organization`  endpoints that needs user authentication  are using the updated context approach.

PR #9, opened 2024-09-09
feature: notifications

lucasmenendez:
User notifications to verify email addresses or invite new admins to an organization. Via SMS and email.

Includes a testing email service. Support for no email service definition on boot allowing to verify users without code.

Now, when a new user registers, the account is created as unverified and the user receives a verification email. A new endpoint has been created to verify a user's account based on the user's email and the code sent.

There are also two new endpoints to request a user password recovery, which sends a code to the email provided to recover the associated password and reset the user password with that code.


db/users.go
```diff
@@ -178,3 +182,26 @@ func (ms *MongoStorage) IsMemberOf(userEmail, organizationAddress string, role U
...
+// user verification code in the verifications collection and returns the user
+// with the ID associated with the verification code.
+func (ms *MongoStorage) UserByVerificationCode(code string, t CodeType) (*User, error) {
+	ms.keysLock.RLock()
+	defer ms.keysLock.RUnlock()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	result := ms.verifications.FindOne(ctx, bson.M{"code": code, "type": t})
```
emmdim, line 198: @lucasmenendez shouldn't this be part of the verifications code?

db/users.go
```diff
@@ -178,3 +182,26 @@ func (ms *MongoStorage) IsMemberOf(userEmail, organizationAddress string, role U
...
+	verification := &UserVerification{}
+	if err := result.Decode(verification); err != nil {
+		if err == mongo.ErrNoDocuments {
+			return nil, ErrNotFound
+		}
+		return nil, err
+	}
+	return ms.user(verification.ID)
```
emmdim, line 206: @lucasmenendez would there be an issue if you would use here the `User`? If that is the case, and together with the verifications comment above, no locks are needed and maybe the `user` helper function is not necessary anymore. What do you think?

db/verifications.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	opts := options.Replace().SetUpsert(true)
+	_, err := ms.verifications.ReplaceOne(ctx, filter, verification, opts)
+	return err
+}
+
+// VerifyUserAccount method verifies the user provided, modifying the user to
+// mark as verified and removing the verification code. If an error occurs, it
+// returns the error.
+func (ms *MongoStorage) VerifyUserAccount(user *User) error {
```
emmdim, line 60: @lucasmenendez shouldn't this be with the user code?

db/verifications.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	// update the user to mark as verified
+	filter := bson.M{"_id": user.ID}
+	if _, err := ms.users.UpdateOne(ctx, filter, bson.M{"$set": bson.M{"verified": true}}); err != nil {
+		return err
+	}
+	// remove the verification code
+	_, err := ms.verifications.DeleteOne(ctx, bson.M{"_id": user.ID, "type": CodeTypeAccountVerification})
```
emmdim, line 75: @lucasmenendez maybe this calla could be a separate function living in this file?

notifications/twilio/sms.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (tsms *TwilioSMS) SendNotification(_ context.Context, notification *notifications.Notification) error {
+	// create message with configured sender number and notification data
+	params := &api.CreateMessageParams{}
+	params.SetTo(notification.ToNumber)
+	params.SetFrom(tsms.config.FromNumber)
+	params.SetBody(notification.Body)
+	// send the message
+	_, err := tsms.client.Api.CreateMessage(params)
```
emmdim, line 56: @lucasmenendez is this call blokcing? If so maybe it should be in a separate goroutine?

api/types.go
```diff
@@ -62,6 +63,18 @@ type UserPasswordUpdate struct {
 	NewPassword string `json:"newPassword"`
 }
 
+// UserVerificationRequest is the request to verify a user.
+type UserVerification struct {
+	Email string `json:"email"`
```
emmdim, line 68: @lucasmenendez should the sms part be included here as well?

cmd/service/main.go
```diff
@@ -66,18 +74,33 @@ func main() {
...
+		apiConf.MailService = new(sendgrid.SendGridEmail)
+		if err := apiConf.MailService.Init(&sendgrid.SendGridConfig{
+			FromName:    sendgridFromName,
+			FromAddress: sendgridFromAddress,
+			APIKey:      sendgridAPIKey,
+		}); err != nil {
+			log.Fatalf("could not create the email service: %v", err)
+		}
+	}
+	log.Infow("email service created", "from", fmt.Sprintf("%s <%s>", sendgridFromName, sendgridFromAddress))
```
emmdim, line 99: @lucasmenendez shoulnd't this print be inside the sendgrid `if`?

emmdim (changes requested):
Additionally please change the random hex for a TOTP as discussed

emmdim (commented):
Additionally please change the random hex for a TOTP as discussed

db/users.go
lucasmenendez: Probably i am not understand your comment... event this method is moved to `db/users.go`, the helper still being required to avoid duplicated code. The helper is private to ensure that it is used under a lock.

notifications/twilio/sms.go
lucasmenendez: It is blocking, just like the sendgrid email `SendNotification` method is. But i dont think that they should be in a goroutine. The caller should know if the method fails, and if they are called concurrently, we need to implement a communication mechanism to know the result.

lucasmenendez: I think it is done now

docker-compose.yml
```diff
@@ -2,8 +2,11 @@ name: saas-backend
 
 services:
   api:
+    platform: linux/amd64
     env_file:
       - .env
+    environment:
+      - VOCDONI_MONGOURL=mongodb://root:vocdoni@mongo:27017/
```
emmdim, line 9: @lucasmenendez doesn't it make sense to import this from .env just as with `VOCDONI_PORT`

api/helpers.go
```diff
@@ -64,6 +54,52 @@ func (a *API) buildLoginResponse(id string) (*LoginResponse, error) {
...
+// sendUserCode method allows to send a code to the user via email or SMS. It
+// generates a verification code and stores it in the database associated to
+// the user email. If the mail service is available, it sends the verification
+// code via email. If the SMS service is available, it sends the verification
+// code via SMS. The code is generated associated a the type of code received,
+// that can be either a verification code or a password reset code. Other types
+// of codes can be added in the future. If neither the mail service nor the SMS
+// service are available, the verification code will be empty but stored in the
+// database to mock the verification process in any case.
+func (a *API) sendUserCode(ctx context.Context, user *db.User, codeType db.CodeType) error {
```
emmdim, line 66: @lucasmenendez it seems that this function is used only by the `api/users.go` module. MAybe it could be moved there?

internal/utils.go
```diff
@@ -0,0 +1,48 @@
+package internal
```
emmdim, line 1: @lucasmenendez why this packages is called internal and not utils?

emmdim (changes requested):

internal/utils.go
lucasmenendez: It is called `internal` because this way go prevents any third party package from importing it, only the packages of the same project ([read more here](https://dev.to/stefanalfbo/internal-packages-in-go-2h9)). 

We can create another package inside the internal one called `utils` if you want, but for me it does not make sense at the moment because it is the only file inside the `internal` package.

api/helpers.go
lucasmenendez: Yes, you're right.

docker-compose.yml
lucasmenendez: No, because it is hardcoded into the other compose services.

PR #13, opened 2024-09-19
feature: extended sign included faucet package in more tx types

lucasmenendez:
The aim of this PR is to extend the tx sign to include faucet package in the SaaS backend for more tx types. In the near future, the inclusion of the faucet package should depend on the type of transaction and the plan of the organization trying to sign it.


api/transaction.go
```diff
@@ -101,19 +101,127 @@ func (a *API) signTxHandler(w http.ResponseWriter, r *http.Request) {
...
+			if txNewProcess == nil || txNewProcess.Process == nil || txNewProcess.Nonce == 0 {
+				ErrInvalidTxFormat.With("missing fields").Write(w)
+				return
+			}
+			// check the tx subtype
+			switch txNewProcess.Txtype {
+			case models.TxType_NEW_PROCESS:
+				// generate a new faucet package if it's not present and include it in the tx
+				if txNewProcess.FaucetPackage == nil {
+					faucetPkg, err := a.account.FaucetPackage(organizationSigner.AddressString(), bootStrapFaucetAmount)
```
p4u, line 116: This doesn't seem right, you need to calculate the cost for this transaction and create the exact faucet package amount. 

You need to import [this package ](https://github.com/vocdoni/vocdoni-node/blob/main/vochain/state/electionprice/electionprice.go) and initialize the ElectionPriceCalculator with the chain values (at this point, we probably need to define a vocdoni-node API endpoint to query for the chain parameters).

api/transaction.go
```diff
@@ -101,19 +101,127 @@ func (a *API) signTxHandler(w http.ResponseWriter, r *http.Request) {
...
+			case models.TxType_SET_PROCESS_DURATION:
+				// check if the process duration is in the tx
+				if txSetProcess.Duration == nil {
+					ErrInvalidTxFormat.With("missing duration field").Write(w)
+					return
+				}
+			}
+			// include the faucet package in the tx if it's not present
+			if txSetProcess.FaucetPackage == nil {
+				faucetPkg, err := a.account.FaucetPackage(organizationSigner.AddressString(), bootStrapFaucetAmount)
```
p4u, line 171: Same here, you need to set the exact price amount for the tx. In the case of these transactions, you should get the price from the vocdoni-node API.

api/transaction.go
```diff
@@ -68,31 +67,39 @@ func (a *API) signTxHandler(w http.ResponseWriter, r *http.Request) {
...
 			// check the tx subtype
 			switch txSetAccount.Txtype {
 			case models.TxType_CREATE_ACCOUNT:
 				// generate a new faucet package if it's not present and include it in the tx
 				if txSetAccount.FaucetPackage == nil {
-					faucetPkg, err := a.account.FaucetPackage(organizationSigner.AddressString(), bootStrapFaucetAmount)
+					// get the tx cost for the tx type
+					amount, ok := a.account.TxCosts[models.TxType_CREATE_ACCOUNT]
+					if !ok {
+						ErrInvalidTxFormat.With("invalid tx type").Write(w)
```
p4u, line 93: this must never happen. So better a panic.

api/transaction.go
```diff
@@ -101,19 +108,203 @@ func (a *API) signTxHandler(w http.ResponseWriter, r *http.Request) {
...
+			}
+			// check the tx subtype
+			switch txNewProcess.Txtype {
+			case models.TxType_NEW_PROCESS:
+				// generate a new faucet package if it's not present and include it in the tx
+				if txNewProcess.FaucetPackage == nil {
+					// get the tx cost for the tx type
+					amount, ok := a.account.TxCosts[models.TxType_NEW_PROCESS]
+					if !ok {
+						ErrInvalidTxFormat.With("invalid tx type").Write(w)
```
p4u, line 126: same here

api/transaction.go
```diff
@@ -101,19 +108,203 @@ func (a *API) signTxHandler(w http.ResponseWriter, r *http.Request) {
...
+					AnonymousVotes:          currentProcess.VoteMode.Anonymous,
+					MaxVoteOverwrite:        currentProcess.TallyMode.MaxVoteOverwrites,
+				})
+			case models.TxType_SET_PROCESS_QUESTION_INDEX:
+				// check if the process question index is in the tx
+				if txSetProcess.QuestionIndex == nil {
+					ErrInvalidTxFormat.With("missing question index field").Write(w)
+					return
+				}
+			case models.TxType_SET_PROCESS_RESULTS:
```
p4u, line 204: we don't want this tx here

api/transaction.go
```diff
@@ -101,19 +108,203 @@ func (a *API) signTxHandler(w http.ResponseWriter, r *http.Request) {
...
+				// faucet package with the required amount for this type of
+				// election update
+				amount += a.account.ElectionPriceCalc.Price(&electionprice.ElectionParameters{
+					MaxCensusSize:           txSetProcess.GetCensusSize(),
+					ElectionDurationSeconds: uint32(currentProcess.EndDate.Sub(currentProcess.StartDate).Seconds()),
+					EncryptedVotes:          currentProcess.VoteMode.EncryptedVotes,
+					AnonymousVotes:          currentProcess.VoteMode.Anonymous,
+					MaxVoteOverwrite:        currentProcess.TallyMode.MaxVoteOverwrites,
+				})
+			case models.TxType_SET_PROCESS_QUESTION_INDEX:
```
p4u, line 198: I think this tx is not even implemented

p4u (approved):
we should have tests for this....

PR #16, opened 2024-09-20
feature: expirable verification codes for user account registration

### LINKED ISSUE #15: implement account verification codes expiration
Currently, the verification codes generated to verify new accounts are valid forever. This behaviour is vulnerable to brute force attacks that attempt to verify an account by trying random codes. To prevent this type of attack, we need to create codes with an expiration date. 

If verification codes can expire, the API should provide some information, such as when the current verification code for an account will expire, or support for resending a verification code when the current one expires.

New API endpoints:
* `GET /users/verify/code?email={emailToVerifyCode}` 
  * **response**
  ```json
  {
    "valid": true,
    "expiration": "2009-11-10T23:00:00Z"
  }
  ```
  
* `POST /users/verify/code`
  * **body**
  ```json
  {
     "email": "{emailToResendCode}"
     // "phone": "+01000000000"
  }
  ```
  * **response**
    * `OK 200`: New code sent to the user.
    * `401 Unauthorized`: The current code associated with this user is still valid.

lucasmenendez:
Closes #15


db/verifications.go
```diff
@@ -46,17 +54,17 @@ func (ms *MongoStorage) UserVerificationCode(user *User, t CodeType) (string, er
...
 	}
-	return verification.Code, nil
+	return verification, nil
 }
 
 // SetVerificationCode method sets the verification code for the user provided.
 // If the user already has a verification code, it updates it. If an error
 // occurs, it returns the error.
-func (ms *MongoStorage) SetVerificationCode(user *User, code string, t CodeType) error {
+func (ms *MongoStorage) SetVerificationCode(user *User, code string, t CodeType, exp time.Time) error {
```
emmdim, line 67: @lucasmenendez maybe a bit late for this comment, but is there a reason that the function waits for `*User` and not a userID ?
Same question applies for `UserVerifiactionCode` function

db/helpers.go
```diff
@@ -129,6 +129,13 @@ func (ms *MongoStorage) createIndexes() error {
 	if _, err := ms.verifications.Indexes().CreateOne(ctx, verificationCodeIndex); err != nil {
 		return fmt.Errorf("failed to create index on code for verifications: %w", err)
 	}
+	// create an index for the 'type' field on user verifications
+	verificationTypeIndex := mongo.IndexModel{
```
emmdim, line 133: @lucasmenendez What does an index to a non unique field mean abstractly? How is this is going to optimize searching?

db/helpers.go
```diff
@@ -129,6 +129,13 @@ func (ms *MongoStorage) createIndexes() error {
 	if _, err := ms.verifications.Indexes().CreateOne(ctx, verificationCodeIndex); err != nil {
```
emmdim, line 129: @lucasmenendez a late comment that I should have caught in a previous PR.
The index is there to help the search. IMHO, since in your searches you use the tuple `(code,type)` then you should either update your index or the search filter.

api/routes.go
```diff
@@ -20,6 +20,9 @@ const (
 	usersEndpoint = "/users"
 	// POST /users/verify to verify the user
 	verifyUserEndpoint = "/users/verify"
+	// GET /users/verify/code to get the user verification code information
+	// POST /users/verify/code to try to resend the user verification code
+	verifyUserCodeEndpoint = "/users/verify/code"
```
emmdim, line 25: @lucasmenendez shouldn't this endpoint have a more obvious name? 
If I understand correctly the fucntion, something like `retry` or similar would make more sense

api/users.go
```diff
@@ -162,6 +199,108 @@ func (a *API) verifyUserAccountHandler(w http.ResponseWriter, r *http.Request) {
...
+// code information of a user. It requires the user email to be provided. It
+// returns the user email, the verification code, the phone number, the code
+// expiration and if the code is valid (not expired and has not reached the
+// maximum number of attempts). If the user is already verified, an error is
+// returned. If the user is not found, an error is returned. If the
+// verification code is not found, an error is returned. If any other error
+// occurs, a generic error is returned.
+func (a *API) userVerificationCodeInfoHandler(w http.ResponseWriter, r *http.Request) {
+	// get the user email from the request query
+	userEmail := r.URL.Query().Get("email")
```
emmdim, line 212: @lucasmenendez is the sms part missing here?

api/users.go
```diff
@@ -162,6 +199,108 @@ func (a *API) verifyUserAccountHandler(w http.ResponseWriter, r *http.Request) {
...
+func (a *API) resendUserVerificationCodeHandler(w http.ResponseWriter, r *http.Request) {
+	verification := &UserVerification{}
+	if err := json.NewDecoder(r.Body).Decode(verification); err != nil {
+		ErrMalformedBody.Write(w)
+		return
+	}
+	if verification.Email == "" {
+		ErrInvalidUserData.With("no email provided").Write(w)
+		return
+	}
```
emmdim, line 265: @lucasmenendez is the sms part missing here?

emmdim (changes requested):

db/verifications.go
lucasmenendez: I think that receives the user struct is more flexible, it does not depends of a single information about the user but also allows to set an `*User` instance with just the `userID`.

db/helpers.go
lucasmenendez: fixed, ill remove this index and modify the index over `code` to be over the tuple `code-type` which is always unique.

api/users.go
lucasmenendez: yes, it will be added in the next commit

api/routes.go
lucasmenendez: It is also used to get the expiration time of an existing verification code, so here is the HTTP verb that adds more information about the endpoint (`GET` to retrieve info about the code and `POST` to create a new code)

emmdim (approved):

PR #14, opened 2024-09-19
feature: html email templates

lucasmenendez:
It includes the support to send email notifications using HTML templates. Now there are 3 templates available:
* Email verification.
* Password recovery.
* Admins invites.

And one more that will be included in the future:
* Welcome (to be sent after account verification).

PR #18, opened 2024-09-24
feature: generic smtp provider for email notifications

lucasmenendez:
This PR replaces the current email notification service from the Sendgrid API with a generic SMTP implementation, allowing the email provider to be changed by simply updating envvars.

Some new code has been commented to be uncommented when #14 is merged.


notifications/smtp/smtp.go
```diff
@@ -0,0 +1,114 @@
...
+	SMTPUsername string
+	SMTPPassword string
+}
+
+type SMTPEmail struct {
+	config *SMTPConfig
+	auth   smtp.Auth
+}
+
+func (se *SMTPEmail) Init(rawConfig any) error {
```
p4u, line 29: Comments please, at least on all exported methods and types.

p4u, line 29: Nit: Use a constructor function instead of a Init() metod.

notifications/smtp/smtp.go
```diff
@@ -0,0 +1,114 @@
...
+		return fmt.Errorf("invalid SMTP configuration")
+	}
+	// parse from email
+	if _, err := mail.ParseAddress(config.FromAddress); err != nil {
+		return fmt.Errorf("could not parse from email: %v", err)
+	}
+	// set configuration in struct
+	se.config = config
+	// init SMTP auth
+	se.auth = smtp.PlainAuth("", se.config.SMTPUsername, se.config.SMTPPassword, se.config.SMTPServer)
```
p4u, line 42: The SMTP server could allow no authentication (i.e for local testing using local sendmail). So auth should be only configured if user/password are defined, and they should not be enforced to exist.

notifications/smtp/smtp.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type SMTPEmail struct {
+	config *SMTPConfig
+	auth   smtp.Auth
+}
+
+func (se *SMTPEmail) Init(rawConfig any) error {
+	// parse configuration
+	config, ok := rawConfig.(*SMTPConfig)
```
p4u, line 31: Umm, why rawConfig is any? Is this implementing an interface?

p4u (changes requested):

notifications/smtp/smtp.go
lucasmenendez: yes it is an interface

notifications/smtp/smtp.go
lucasmenendez: Updated with new methods that are also required for testing.

p4u (approved):

PR #21, opened 2024-10-02
feature: strip vochain data

lucasmenendez:
remove organization data that is also provided by the sdk and vochain


emmdim (commented):
LGTM need to do some E2E end testing before appriving

PR #24, opened 2024-10-31
feature: new endpoint for organization admins to get pending invitations

lucasmenendez:
check the updated [api docs](https://github.com/vocdoni/saas-backend/blob/f/pending_org_invititations/api/docs.md#-list-pending-invitations).

### DIFF
--- a/api/docs.md
+++ b/api/docs.md
@@ -27,6 +27,7 @@
   - [🔍 Organization info](#-organization-info)
   - [🧑‍🤝‍🧑 Organization members](#-organization-members)
   - [🧑‍💼 Invite organization member](#-invite-organization-member)
+  - [⏳ List pending invitations](#-list-pending-invitations)
   - [🤝 Accept organization invitation](#-accept-organization-invitation)
   - [🤠 Available organization members roles](#-available-organization-members-roles)
   - [🏛️ Available organization types](#-available-organization-types)
@@ -575,6 +576,8 @@ Only the following parameters can be changed. Every parameter is optional.
 
 * **Path** `/organizations/{address}/members`
 * **Method** `POST`
+* **Headers**
+  * `Authentication: Bearer <user_token>`
 * **Request**
 ```json
 {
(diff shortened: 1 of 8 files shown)


api/docs.md
```diff
@@ -598,6 +601,35 @@ Only the following parameters can be changed. Every parameter is optional.
 | `409` | `40901` | `duplicate conflict` |
 | `500` | `50002` | `internal server error` |
 
+### ⏳ List pending invitations
+
+* **Path** `/organizations/{address}/members/pending`
+* **Method** `POST`
```
elboletaire, line 607: Isn't it GET?

elboletaire (changes requested):

api/docs.md
lucasmenendez: yes, it is

elboletaire (approved):

api/routes.go
```diff
@@ -50,6 +50,8 @@ const (
 	organizationAddMemberEndpoint = "/organizations/{address}/members"
 	// POST /organizations/{address}/members/invite/accept to accept the invitation
 	organizationAcceptMemberEndpoint = "/organizations/{address}/members/accept"
+	// GET /organizations/{address}/members/pending to get the pending members
+	organizationPendingMemberEndpoint = "/organizations/{address}/members/pending"
```
emmdim, line 54: @lucasmenendez  organizationPendingMember{s}Endpoint i gues :-P

api/organizations.go
```diff
@@ -397,6 +397,45 @@ func (a *API) acceptOrganizationMemberInvitationHandler(w http.ResponseWriter, r
...
+	if !ok {
+		ErrNoOrganizationProvided.Write(w)
+		return
+	}
+	if !user.HasRoleFor(org.Address, db.AdminRole) {
+		ErrUnauthorized.Withf("user is not admin of organization").Write(w)
+		return
+	}
+	// check if the user is already verified
+	if !user.Verified {
```
emmdim, line 418: @lucasmenendez  why is that relevenat here?

api/routes.go
lucasmenendez: yep, i gues{s} :P

api/organizations.go
lucasmenendez: to prevent that not verified users to perform actions

emmdim: But what is the action in this case? I mean you have check that the user is Admin, and your a getting the pending invitations...

lucasmenendez: The operation is to get a list of pending invitations, and yes, I've already checked that the user has an admin role for the requested org, but I think if the account is not yet verified, the user should not perform any operation, not even list invitations.

emmdim:
LGTM

api/organizations.go
emmdim: @lucasmenendez will create new PR that will improve the authorization/verifcation  helpers of the API

PR #26, opened 2024-11-06
feature: improve and extend auth middleware

### LINKED ISSUE #25: improve and extend auth middleware
1. Create a helper with the logic of the current auth middleware.
2. Modify the auth middleware to use the helper and also ensure that the user account is already verified.
3. Move the verify endpoint to the public endpoints and use the helper on that endpoint directly, instead of the middleware.

lucasmenendez:
After code review, there is no need to implement a helper or make the verification endpoint public, because it already is. Simply extend the current auth middleware to enforce that user accounts have been verified in order to perform any operation that requires it. 

It closes #25


emmdim (approved):
LGTM

PR #29, opened 2024-11-19
 Reads plans from stripe instead from local file

emmdim:
* Uses local file for testing


lucasmenendez (changes requested):

stripe/stripe.go
```diff
@@ -57,3 +66,73 @@ func (s *StripeClient) GetInfoFromEvent(event stripe.Event) (*stripe.Customer, *
...
 }
+
+func (s *StripeClient) GetPriceByID(priceID string) *stripe.Price {
+	params := &stripe.PriceSearchParams{
+		SearchParams: stripe.SearchParams{
+			Query: fmt.Sprintf("active:'true' AND lookup_key:'%s'", priceID),
+		},
+	}
+	params.AddExpand("data.tiers")
+	results := price.Search(params)
```
lucasmenendez, line 77: also:

```go
if results := price.Search(params); results.Next() {
	return results.Price()
}
```

lucasmenendez (changes requested):
as steve jobs would say: one more thing

lucasmenendez (approved):

PR #31, opened 2024-12-11
Store and retrieve files (for now images) to an s3 like object storage

emmdim:
## Endpoints
Currently one private endpoints will be enabled:
   *  `/upload`
The download endpoint is for now ommited since the images will be retrieved directly from the  storage hosting service cdn
 
 ## Key, naming and buckets
There will be a separate route regarding org metadata images and another one for election metadata images
 * organization metadata images:
     *  `organization/${sha(imageData)}.{png | jpg | jpeg}`
*  election metadata images:
     * `election/${sha(imageData)}{png | jpg | jpeg}`  
## Notes:
* The service is compativle with whichever service can be used through the s3 sdk v2.


objectstorage/objectstorage.go
```diff
@@ -0,0 +1,92 @@
...
+		config: config,
+		client: minioClient,
+		bucket: &minio.BucketInfo{
+			Name: apiBucket,
+		},
+	}, nil
+}
+
+// Put uploads a file to the object storage service.
+func (osc *ObjectStorageClient) Put(
```
lucasmenendez, line 50: Few suggestions:
* The comment is quite short for this method, I miss some information about the inputs and their types.
* I suggest that the object can be included into a struct to wrap the `key`, `type`, `size` and `metadata`. It can also include the content when it is returned, for example as `Get` method result.
* If `contentDisposition` has some possible values, they can be public constans or even create multiple `Put` methods: `PutInline`, ...

emmdim:
Replaced by #34

PR #33, opened 2024-12-19
stripe subscription E2E payment integration

emmdim:
This commit represents a significant integration of Stripe's subscription payment system, allowing organizations to manage their subscriptions through Stripe's checkout process while maintaining synchronization with the application's database.

# Key changes
## API Routes & Handlers:
   * Added new endpoints for Stripe integration:
       * POST /subscriptions/checkout - Creates new subscription checkout sessions
       * GET /subscriptions/checkout/{sessionID} - Retrieves checkout session information
   * Enhanced webhook handling to process subscription events of 
POST /subscriptions/webhook
      * customer.subscription.created
      * customer.subscription.updated
      * customer.subscription.deleted
      
## Database Schema Updates
   * Modified OrganizationSubscription structure:
        * Removed EndDate field
   * Updated Plan structure:
         * Added StripePriceID field

## Error Handling:
* Added new error definition: ErrStripeError

### DIFF
--- a/api/docs.md
+++ b/api/docs.md
@@ -33,8 +33,11 @@
   - [🤠 Available organization members roles](#-available-organization-members-roles)
   - [🏛️ Available organization types](#-available-organization-types)
 - [🏦 Plans](#-plans)
-  - [🛒 Get Available Plans](#-get-plans)
-  - [🛍️ Get Plan Info](#-get-plan-info)
+  - [📋 Get Available Plans](#-get-plans)
+  - [📄 Get Plan Info](#-get-plan-info)
+- [🔰 Subscriptions](#-subscriptions)
+  - [🛒 Create Checkout session](#-create-checkout-session)
+  - [🛍️ Get Checkout session info](#-get-checkout-session-info)
 
 </details>
 
@@ -324,8 +327,7 @@ This endpoint only returns the addresses of the organizations where the current 
         "subscription":{
             "PlanID":3,
             "StartDate":"2024-11-07T15:25:49.218Z",
-            "EndDate":"0001-01-01T00:00:00Z",
-            "RenewalDate":"0001-01-01T00:00:00Z",
+            "RenewalDate":"2025-11-07T15:25:49.218Z",
             "Active":true,
             "MaxCensusSize":10
         },
@@ -913,6 +915,60 @@ This request can be made only by organization admins.
 
 | HTTP Status | Error code | Message |
 |:---:|:---:|:---|
+| `400` | `40004` | `malformed JSON body` |
+| `404` | `40009` | `plan not found` |
+| `500` | `50001` | `internal server error` |
+
+
+## 🔰 Subscriptions
+
+### 🛒 Create Checkout session
+
+* **Path** `/subscriptions/checkout/`
+* **Method** `POST`
+* **Request Body** 
+```json
+{
+  "lookupKey": 1, // PLan's corresponging DB ID
+  "returnURL": "https://example.com/return",
+  "address": "user@mail.com",
+  "amount": 1000, // The desired maxCensusSize
+}
+```
+
+* **Response**
+```json
+{
+  "id": "cs_test_a1b2c3d4e5f6g7h8i9j0",
+   // ... rest of stripe session attributes
+}
+```
+
+* **Errors**
+
+| HTTP Status | Error code | Message |
+|:---:|:---:|:---|
 | `400` | `40010` | `malformed URL parameter` |
 | `400` | `40023` | `plan not found` |
+| `500` | `50002` | `internal server error` |
+
+### 🛍️ Get Checkout session info
+
+* **Path** `/subscriptions/checkout/{sessionID}`
+* **Method** `GET`
+* **Response**
+```json
+{
+  "status": "complete", // session status
+  "customer_email": "customer@example.com",
+  "subscription_status": "active"
+}
+```
+
+* **Errors**
+
+| HTTP Status | Error code | Message |
+|:---:|:---:|:---|
+| `400` | `40010` | `malformed URL parameter` |
+| `400` | `40023` | `session not found` |
 | `500` | `50002` | `internal server error` |
--- a/api/stripe.go
+++ b/api/stripe.go
@@ -1,25 +1,32 @@
 package api
 
 import (
+	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"time"
 
+	"github.com/go-chi/chi/v5"
+	"github.com/stripe/stripe-go/v81"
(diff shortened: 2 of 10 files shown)


api/docs.md
```diff
@@ -35,6 +35,9 @@
 - [🏦 Plans](#-plans)
   - [🛒 Get Available Plans](#-get-plans)
   - [🛍️ Get Plan Info](#-get-plan-info)
+- [Subscriptions](#-subscriptions)
```
lucasmenendez, line 38: to be completed

api/stripe.go
```diff
@@ -84,6 +86,120 @@ func (a *API) handleWebhook(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		log.Debugf("stripe webhook: subscription %s for organization %s processed successfully", subscription.ID, org.Address)
+	case "customer.subscription.updated", "customer.subscription.deleted":
```
lucasmenendez, line 89: why are both events handled in the same case and then it contains an if to handle them in a different way?

api/types.go
```diff
@@ -183,3 +183,11 @@ type OrganizationSubscriptionInfo struct {
 	Usage              *db.OrganizationCounters     `json:"usage"`
 	Plan               *db.Plan                     `json:"plan"`
 }
+
+type SubscriptionCheckout struct {
```
lucasmenendez, line 187: Please, add a comment to this struct to give more context to the information that it contains.

lucasmenendez (changes requested):
The code looks good and is well structured. One suggestion, to improve clarity would be to add some comments in the code, not just in the functions and structs declarations, but also around the more complex parts of the logic or third party code.

api/stripe.go
emmdim: becaise there os a lot of repeated code. The other way would be to create helpr functions. @lucasmenendez do you think it makes more sense?

lucasmenendez: Probably creating some helpers could be the way, at least for these operations, which are common to all events:

```go
customer, subscription, err := a.stripe.GetInfoFromEvent(*event)
if err != nil {
	log.Errorf("stripe webhook: error getting info from event: %s\n", err.Error())
	w.WriteHeader(http.StatusBadRequest)
	return
}
address := subscription.Metadata["address"]
if len(address) == 0 {
	log.Errorf("subscription %s does not contain an address in metadata", subscription.ID)
	w.WriteHeader(http.StatusBadRequest)
	return
}
org, _, err := a.db.Organization(address, false)
if err != nil || org == nil {
	log.Errorf("could not update subscription %s, a corresponding organization with address %s was not found.",
		subscription.ID, address)
	log.Errorf("please do manually for creator %s \n  Error:  %s", customer.Email, err.Error())
	w.WriteHeader(http.StatusBadRequest)
	return
}
```

api/docs.md
lucasmenendez: `🔰 Subscriptions`

api/stripe.go
```diff
@@ -35,55 +41,180 @@ func (a *API) handleWebhook(w http.ResponseWriter, r *http.Request) {
...
+
+	httpWriteJSON(w, status)
+}
+
+// getSubscriptionOrgInfo is a helper function that retrieves the subscription information from
+// the subscription event and the Organization information from the database.
+func (a *API) getSubscriptionOrgInfo(event *stripe.Event) (*stripeService.StripeSubscriptionInfo, *db.Organization, error) {
+	stripeSubscriptionInfo, err := a.stripe.GetSubscriptionInfoFromEvent(*event)
+	if err != nil {
+		return &stripeService.StripeSubscriptionInfo{}, &db.Organization{}, err
```
lucasmenendez, line 213: just return nil for pointers with the error

api/stripe.go
```diff
@@ -35,55 +41,180 @@ func (a *API) handleWebhook(w http.ResponseWriter, r *http.Request) {
...
+	}
+	org, _, err := a.db.Organization(stripeSubscriptionInfo.CustomerEmail, false)
+	if err != nil || org == nil {
+		return &stripeService.StripeSubscriptionInfo{}, &db.Organization{}, err
```
lucasmenendez, line 217: here `err` can be nil, manage each case independently instead

api/stripe.go
```diff
@@ -35,55 +41,180 @@ func (a *API) handleWebhook(w http.ResponseWriter, r *http.Request) {
...
+		stripeSubscriptionInfo, org, err := a.getSubscriptionOrgInfo(event)
 		if err != nil || org == nil {
 			log.Errorf("could not update subscription %s, a corresponding organization with address %s was not found.",
-				subscription.ID, address)
-			log.Errorf("please do manually for creator %s \n  Error:  %s", customer.Email, err.Error())
+				stripeSubscriptionInfo.ID, stripeSubscriptionInfo.OrganizationAddress)
+			log.Errorf("please do manually for creator %s \n  Error:  %s", stripeSubscriptionInfo.CustomerEmail, err.Error())
 			w.WriteHeader(http.StatusBadRequest)
 			return
 		}
```
lucasmenendez, line 52: This code could be included in `getSubscriptionOrgInfo` helper, at least return an error when org is nil because the check is done every time that that function is executed.

lucasmenendez (changes requested):

stripe/stripe.go
```diff
@@ -13,13 +15,32 @@ import (
...
 }
 
+// ReturnStatus represents the response structure for checkout session status
+type ReturnStatus struct {
+	Status             string `json:"status"`
+	CustomerEmail      string `json:"customer_email"`
+	SubscriptionStatus string `json:"subscription_status"`
+}
+
+type StripeSubscriptionInfo struct {
```
lucasmenendez, line 33: to be commented

api/stripe.go
```diff
@@ -35,55 +42,186 @@ func (a *API) handleWebhook(w http.ResponseWriter, r *http.Request) {
...
+
+	httpWriteJSON(w, status)
+}
+
+// getSubscriptionOrgInfo is a helper function that retrieves the subscription information from
+// the subscription event and the Organization information from the database.
+func (a *API) getSubscriptionOrgInfo(event *stripe.Event) (*stripeService.StripeSubscriptionInfo, *db.Organization, error) {
+	stripeSubscriptionInfo, err := a.stripe.GetSubscriptionInfoFromEvent(*event)
+	if err != nil {
+		return nil, &db.Organization{}, fmt.Errorf("could not decode event for subscription: %s", err.Error())
```
lucasmenendez, line 211: `nil` instead pointer to `db.Organization`

lucasmenendez (approved):
some nits

PR #34, opened 2024-12-20
Implements image storage

emmdim:
## Overview

Iimplements an image upload/download system with the following key features:

- Complete image upload and download API endpoints with authentication
- Integration with an object storage backend (likely S3 or similar)
- Database support for tracking uploaded objects
- User-specific file management
- Multipart form handling for file uploads
- Proper content type handling for downloads
- Error handling for various upload/download scenarios

The implementation suggests a robust file storage system that will be used for handling user-uploaded images, with proper integration into the existing authentication and user management system.

### Added Files
- **api/object_storage.go**: Implements new image upload and download endpoints
- **db/object.go**: Database layer for object storage
- **objectstorage/objectstorage.go**: Core object storage client implementation

### New Endpoints

1. **Image Upload Endpoint**
   - Handler: `uploadImageWithFormHandler`
   - Features:
     - Supports multipart form file uploads
     - Maximum file size: 32MB
     - Requires user authentication
     - Returns URLs of uploaded files
     - Handles multiple file uploads in a single request
     - Associates uploads with user email

2. **Image Download Endpoint**
   - Handler: `downloadImageInlineHandler`
   - Features:
     - Retrieves images by object ID
     - Serves images inline with appropriate content type
     - Handles content disposition for browser display


api/object_storage.go
```diff
@@ -0,0 +1,73 @@
...
+			ErrGenericInternalServerError.Withf("cannot upload file %s", err.Error()).Write(w)
+			break
+		}
+		returnURLs = append(returnURLs, url)
+	}
+	httpWriteJSON(w, map[string][]string{"urls": returnURLs})
+}
+
+func (a *API) downloadImageInlineHandler(w http.ResponseWriter, r *http.Request) {
+	objectID := chi.URLParam(r, "objectName")
```
lucasmenendez, line 54: I think that `"objectName"` could be a constant declared in the current file.

api/plans.go
```diff
@@ -20,7 +22,7 @@ func (a *API) getPlansHandler(w http.ResponseWriter, r *http.Request) {
 
 func (a *API) planInfoHandler(w http.ResponseWriter, r *http.Request) {
 	// get the plan ID from the URL
-	planID := r.URL.Query().Get("planID")
+	planID := chi.URLParam(r, "planID")
```
lucasmenendez, line 25: Probably `"plainID"` could also be a constant.

db/object.go
```diff
@@ -0,0 +1,68 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+func (ms *MongoStorage) Object(id string) (*Object, error) {
```
lucasmenendez, line 13: to be commented

objectstorage/objectstorage.go
```diff
@@ -0,0 +1,123 @@
...
+	DB        *db.MongoStorage
+	ServerURL string
+}
+
+type ObjectStorageClient struct {
+	db        *db.MongoStorage
+	serverURL string
+}
+
+var isObjectNameRgx = regexp.MustCompile(`^([a-zA-Z0-9]+)\.(jpg|jpeg|png)`)
```
lucasmenendez, line 25: why does the regex capture the extension of the object instead of just the objectID? i cannot find where the file extension is extracted and used with this regex.

objectstorage/objectstorage.go
```diff
@@ -0,0 +1,123 @@
...
+	return object, nil
+}
+
+// uploadObject uploads the object image with the given objectID, associated to
+// the user with the given userFID and the community with the given communityID.
+// If the objectID is empty, it calculates the objectID from the data. It returns
+// the URL of the uploaded object image. It stores the object in the database.
+// If an error occurs, it returns an empty string and the error.
+func (osc *ObjectStorageClient) Put(data io.Reader, size int64, userID string) (string, error) {
+	// if !isBase64Image(data) {
```
lucasmenendez, line 66: to be removed?

objectstorage/objectstorage.go
```diff
@@ -0,0 +1,123 @@
...
+	// 	return "", fmt.Errorf("image is not base64 encoded")
+	// }
+
+	// Get the expected size of the data
```
lucasmenendez, line 70: to be done?

objectstorage/objectstorage.go
```diff
@@ -0,0 +1,123 @@
...
+	// Create a buffer of the appropriate size
+	buff := make([]byte, size)
+	_, err := data.Read(buff)
+	if err != nil {
+		return "", fmt.Errorf("cannot read file %s", err.Error())
+	}
+	// checking the content type
+	// so we don't allow files other than images
+	filetype := http.DetectContentType(buff)
+	if filetype != "image/jpeg" && filetype != "image/png" && filetype != "image/jpg" {
```
lucasmenendez, line 81: The supported types could be an alias type to string and some constants, and there could be a public variable with a map of the default ones. Then the `ObjectStorageConfig' struct could contain a configurable attribute to define the supported types.

objectstorage/objectstorage.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/vocdoni/saas-backend/db"
+)
+
+type ObjectStorageConfig struct {
+	DB        *db.MongoStorage
+	ServerURL string
+}
+
+type ObjectStorageClient struct {
```
lucasmenendez, line 20: I think we should remove any reference to the API or server urls from this package and move all that logic there. The API already has the route definitions and can encode and decode the resource URL in a more consistent way.

objectstorage/objectstorage.go
```diff
@@ -0,0 +1,123 @@
...
+
+// objectIDfromURL returns the objectID from the given URL. If the URL is not an
+// object URL, it returns an empty string and false.
+func objectIDfromName(url string) (string, bool) {
+	objectID := isObjectNameRgx.FindStringSubmatch(url)
+	if len(objectID) != 3 {
+		return "", false
+	}
+	return objectID[1], true
+}
```
lucasmenendez, line 111: I think that these functions should be placed in the `api` package.

db/helpers.go
```diff
@@ -176,6 +180,14 @@ func (ms *MongoStorage) createIndexes() error {
...
 		return fmt.Errorf("failed to create index on invitationCode for organization invites: %w", err)
 	}
+	// add objectID index to objects collection
+	// objectIDIndex := mongo.IndexModel{
+	// 	Keys:    bson.D{{Key: "_id", Value: 1}}, // 1 for ascending order
+	// 	Options: options.Index().SetUnique(true),
+	// }
+	// if _, err := ms.objects.Indexes().CreateOne(ctx, objectIDIndex); err != nil {
+	// 	return fmt.Errorf("failed to create index on _id for objects: %w", err)
+	// }
```
lucasmenendez, line 190: to be uncommented?

api/users.go
```diff
@@ -174,7 +175,7 @@ func (a *API) verifyUserAccountHandler(w http.ResponseWriter, r *http.Request) {
 // returned.
 func (a *API) userVerificationCodeInfoHandler(w http.ResponseWriter, r *http.Request) {
 	// get the user email of the user from the request query
-	userEmail := r.URL.Query().Get("email")
+	userEmail := chi.URLParam(r, "email")
```
lucasmenendez, line 178: `"email"` could be a constant

cmd/service/main.go
```diff
@@ -12,14 +12,18 @@ import (
...
 	"github.com/vocdoni/saas-backend/notifications/mailtemplates"
 	"github.com/vocdoni/saas-backend/notifications/smtp"
+	"github.com/vocdoni/saas-backend/objectstorage"
 	"github.com/vocdoni/saas-backend/stripe"
 	"github.com/vocdoni/saas-backend/subscriptions"
 	"go.vocdoni.io/dvote/apiclient"
 	"go.vocdoni.io/dvote/log"
 )
 
+var serverURL = "http://localhost:8080"
```
lucasmenendez, line 22: why is it a variable instead just the default value of the flag?

lucasmenendez (changes requested):

api/docs.md
```diff
@@ -916,3 +918,37 @@ This request can be made only by organization admins.
...
+## Storage
+
+### Upload image with origin 
+
+* **Path** `/storage/{origin}`
+* **Method** `POST`
+
+Accepting files uploaded by forms as such:
+```html
+<form action="http://localhost:8000" method="post" enctype="multipart/form-data">
```
lucasmenendez, line 932: Should the `action` URL end with `/storage`?

objectstorage/objectstorage.go
```diff
@@ -0,0 +1,123 @@
+package objectstorage
```
lucasmenendez, line 1: I think that this package should have some errors declared as public to be managed by other package like the `api`. For example those used if the file does not exists, the file type is not supported or the objectID cannot be calculated.

api/object_storage.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		defer func() {
+			if err := file.Close(); err != nil {
+				ErrGenericInternalServerError.Withf("cannot close file %s", err.Error()).Write(w)
+				return
+			}
+		}()
+		// upload the file using the object storage client
+		// and get the URL of the uploaded file
+		url, err := a.objectStorage.Put(file, fileHeader.Size, user.Email)
```
lucasmenendez, line 43: if specific errors were returned, more detailed http statuses could be returned.

lucasmenendez (changes requested):

objectstorage/objectstorage.go
emmdim: It is passed as a function argument after all

api/object_storage.go
emmdim: Hm, not sure I get you. This is not a technique used for the rest of the URI parameters. WHy should we apply it here?

api/plans.go
emmdim: Hm, not sure I get you. This is not a technique used for the rest of the URI parameters. WHy should we apply it here?

api/users.go
emmdim: Hm, not sure I get you. This is not a technique used for the rest of the URI parameters. WHy should we apply it here?

objectstorage/objectstorage.go
emmdim: The regex captures both, but only the obejctID is being used.
The file type is retrieved from the file itself, instead of using the extension provided, the file

db/helpers.go
emmdim: AFAIK the `_id` is automatically indexed by mongo, therefore will remove these lines

api/object_storage.go
lucasmenendez: I think that is better to have all this strings in a constants file in the same package, no only if they are used multiple times, but also for maintainability.

objectstorage/objectstorage.go
lucasmenendez: If it is not being used, then simply match it, but why capture it?

lucasmenendez (approved):

PR #40, opened 2025-01-22
 Implements subscription limitations

emmdim:
* Separate limitations as DB and TX limitations
* Adds handler for SET_ACCOUNT_INFO_URI transaction
* Related to https://github.com/vocdoni/interoperability/issues/240

### DIFF
--- a/db/const.go
+++ b/db/const.go
@@ -5,6 +5,7 @@ const (
 	AdminRole   UserRole = "admin"
 	ManagerRole UserRole = "manager"
 	ViewerRole  UserRole = "viewer"
+	AnyRole     UserRole = "any"
 	// organization types
 	AssemblyType               OrganizationType = "assembly"
 	AssociationType            OrganizationType = "association"
@@ -31,6 +32,7 @@ var writableRoles = map[UserRole]bool{
 	AdminRole:   true,
 	ManagerRole: true,
 	ViewerRole:  false,
+	AnyRole:     false,
 }
 
 // UserRoleNames is a map that contains the user role names by role
@@ -38,6 +40,7 @@ var UserRolesNames = map[UserRole]string{
 	AdminRole:   "Admin",
 	ManagerRole: "Manager",
 	ViewerRole:  "Viewer",
+	AnyRole:     "Any",
 }
 
 // HasWriteAccess function checks if the user role has write access
@@ -81,6 +84,7 @@ var validRoles = map[UserRole]bool{
 	AdminRole:   true,
 	ManagerRole: true,
 	ViewerRole:  true,
+	AnyRole:     true,
 }
 
 // IsValidUserRole function checks if the user role is valid
--- a/db/const.go
+++ b/db/const.go
@@ -84,7 +84,7 @@ var validRoles = map[UserRole]bool{
 	AdminRole:   true,
 	ManagerRole: true,
 	ViewerRole:  true,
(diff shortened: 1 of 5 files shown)


subscriptions/subscriptions.go
```diff
@@ -29,23 +56,134 @@ func New(conf *SubscriptionsConfig) *Subscriptions {
 	}
 }
 
-// HasPermission checks if the organization has permission to perform the given transaction.
-func (p *Subscriptions) HasPermission(
+// HasOrgPermission checks if the organization has permission to perform the given action.
```
lucasmenendez, line 59: It seems that the comment is not updated

db/types.go
```diff
@@ -32,6 +32,15 @@ func (u *User) HasRoleFor(address string, role UserRole) bool {
 	return false
 }
 
+func (u *User) IsMemberOf(address string) bool {
```
lucasmenendez, line 35: I think that it should be commented and moved to `user.go`

subscriptions/subscriptions.go
```diff
@@ -29,23 +56,134 @@ func New(conf *SubscriptionsConfig) *Subscriptions {
...
 	case models.TxType_NEW_PROCESS, models.TxType_SET_PROCESS_CENSUS:
+		// check if the user has the admin role for the organization
+		if !user.HasRoleFor(org.Address, db.AdminRole) {
+			return false, fmt.Errorf("user does not have admin role")
+		}
 		newProcess := tx.GetNewProcess()
 		if newProcess.Process.MaxCensusSize > uint64(org.Subscription.MaxCensusSize) {
 			return false, fmt.Errorf("MaxCensusSize is greater than the allowed")
 		}
+		if org.Counters.Processes+1 > plan.Organization.MaxProcesses {
```
lucasmenendez, line 123: Why not use `>=` instead of add one? I think that is more easy to read.

subscriptions/subscriptions.go
```diff
@@ -29,23 +56,134 @@ func New(conf *SubscriptionsConfig) *Subscriptions {
...
+		if !user.HasRoleFor(org.Address, db.AdminRole) || !user.HasRoleFor(org.Address, db.ManagerRole) {
+			return false, fmt.Errorf("user does not have admin role")
+		}
+
 	}
+
 	return true, nil
 }
+
+func (p *Subscriptions) HasDBPersmission(userEmail, orgAddress string, permission DBPermission) (bool, error) {
```
lucasmenendez, line 140: Please include a comment

lucasmenendez (changes requested):

db/types.go
emmdim: This function is very similar to the function `HasRoleFor`: It does not use the DB just the properties of the structs. Therefore both functions need to be in the same place. If you believe this is user.go, then I will move both of them

db/const.go
```diff
@@ -81,6 +84,7 @@ var validRoles = map[UserRole]bool{
 	AdminRole:   true,
 	ManagerRole: true,
 	ViewerRole:  true,
+	AnyRole:     true,
```
lucasmenendez, line 87: I think that `AnyRole` shouldn't be included in valid roles. If it is included a user can be included in a organization with this role. As far as i know, `AnyRole` should be only for internal use.

lucasmenendez (approved):

PR #44, opened 2025-02-25
Fix census endpoints bugs

lucasmenendez:
* Remove double response on add participants
* New struct to wrap census ID after creation
* API docs updated


emmdim (approved):
LGTM

PR #45, opened 2025-02-25
Adds 2fa service

emmdim:


twofactor/saltedkey.go
```diff
@@ -0,0 +1,126 @@
...
+	// SaltSize is the size of the salt used for derive the new key
+	SaltSize = 20
+)
+
+// SaltedKey is a wrapper around ECDSA and ECDSA Blind that helps signing
+// messages with a known Salt. The Salt is added to the private key curve
+// point in order to derive a new deterministic signing key.
+// The same operation must be perform on the public key side in order to
+// verify the signed messages.
+type SaltedKey struct {
```
p4u, line 27: Assure you understand what the SaltedKey is and how it works (https://developer.vocdoni.io/protocol/census/off-chain-csp#blind-signatures). AFAIK we are not using SaltedKeys here.

p4u (commented):
In general you should put attention on the types and use them correctly.

twofactor/queue.go
```diff
@@ -0,0 +1,100 @@
...
+
+type Queue struct {
+	queue         *goconcurrentqueue.FIFO
+	ttl           time.Duration
+	throttle      time.Duration
+	sendChallenge []SendChallengeFunc
+	response      chan (challengeData)
+}
+
+func newQueue(ttl, throttle time.Duration, sChFns []SendChallengeFunc) *Queue {
```
lucasmenendez, line 34: I think that it should receive the notification service to use, instead an array of functions

twofactor/twofactor.go
```diff
@@ -0,0 +1,359 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	notif := &notifications.Notification{
+		ToNumber:  phone,
+		Subject:   "Vocdoni verification code",
+		PlainBody: fmt.Sprintf("Your authentication code is %s", challenge),
+	}
+	// return tf.notificationServices.Mail.SendNotification(ctx, notif)
+	return sn.SmsNotificationService.SendNotification(ctx, notif)
+}
```
lucasmenendez, line 111: I think that it can be generalized to:

```go
type ChallengeType string

const (
	SMSChallenge ChallengeType = "sms"
	EmailChallenge ChallengeType = "email"
)

type NotificationChallenge struct {
	UserID     internal.HexBytes
	ElectionID internal.HexBytes
	Notification *notification.Notification
	Type ChallengeType
	Retries int
	Success bool
}

func (nc *NotificationChallenge) Send(service notifications.NotificationService) error {
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()
	if err := service.Send(ctx, nc.Notification); err != nil {
		return err
	}
	nc.Success = true
	return nil
}

func NewNotificationChallenge(challengeType ChallengeType, userID, electionID internal.HexBytes, contact, optCode string) (*NotificationChallenge, error) {
	n := &notifications.Notification{
		Subject:   "Vocdoni verification code",
		PlainBody: fmt.Sprintf("Your authentication code is %s", challenge),
	}
	switch challengeType {
	case EmailChallenge:
		n.ToAddress = contact
	case SMSChallenge:
		n.ToPhone = contact
	default:
		return nil, fmt.Errof("invalid notification type")
	}
	return &NotificationChallenge{
		UserID: userID,
		ElectionID: electionID,
		Notification: n,
		Type: challengeType,
	}, nil
}
```

lucasmenendez: In this way you can create the queue with the desired notification service and enqueue the NotificationChallenge struct itself, simplifying the `TwoFactor` and the `Queue` logic. The `Queueu` only needs to call notificationChallenge.Send() with its notification service.

twofactor/twofactor.go
```diff
@@ -0,0 +1,359 @@
...
+		maxAttempts,
+		coolDownTime,
+	); err != nil {
+		return nil, err
+	}
+
+	tf.smsQueue = newQueue(
+		coolDownTime,
+		throttleTime,
+		[]SendChallengeFunc{NewSmsNotifcation(conf.NotificationServices.SMS).SendChallenge},
```
lucasmenendez, line 164: with my approach, only needs to pass the sms notification service

twofactor/twofactor.go
```diff
@@ -0,0 +1,359 @@
...
+	)
+	go tf.smsQueue.run()
+	go tf.queueController(tf.smsQueue)
+
+	tf.mailQueue = newQueue(
+		coolDownTime,
+		throttleTime,
+		[]SendChallengeFunc{NewMailNotifcation(conf.NotificationServices.Mail).SendChallenge},
```
lucasmenendez, line 172: with my approach, only needs to pass the email notification service

twofactor/queue.go
```diff
@@ -0,0 +1,100 @@
...
+
+type challengeData struct {
+	userID     internal.HexBytes
+	electionID internal.HexBytes
+	contact    string
+	challenge  string
+	startTime  time.Time
+	retries    int
+	success    bool
+}
```
lucasmenendez, line 20: With my approach this is replaced by `NotificationChallenge` struct

twofactor/queue.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (c challengeData) String() string {
+	return fmt.Sprintf("%s[%s]", c.contact, c.challenge)
+}
+
+type Queue struct {
+	queue         *goconcurrentqueue.FIFO
+	ttl           time.Duration
+	throttle      time.Duration
+	sendChallenge []SendChallengeFunc
```
lucasmenendez, line 30: With my approach this is no longer needed an should be replaced by the `notification.NotificationService` interface

twofactor/twofactor.go
```diff
@@ -0,0 +1,359 @@
...
+		log.Infof("user %s challenged with %s at contact %s", userID.String(), otpCode, contact)
+	} else if notifType == notifications.SMS {
+		if err := tf.smsQueue.add(userID, electionID, contact, otpCode); err != nil {
+			log.Errorf("cannot enqueue challenge: %v", err)
+			return AuthResponse{Error: "problem with SMS challenge system"}
+		}
+		log.Infof("user %s challenged with %s at contact %s", userID.String(), otpCode, contact)
+	} else {
+		return AuthResponse{Error: "invalid notification type"}
+	}
```
lucasmenendez, line 314: With my approach you just need to call `NewNotificationChallege` with the type and the data and then enqueue it.

lucasmenendez (commented):
Here are some suggestions. I will continue with the review of the rest of the code.

api/process.go
```diff
@@ -56,8 +74,21 @@ func (a *API) createProcessHandler(w http.ResponseWriter, r *http.Request) {
 		ErrGenericInternalServerError.WithErr(err).Write(w)
 		return
 	}
+	orgParticipants, err := a.db.OrgParticipantsMemberships(pubCensus.Census.OrgAddress, pubCensus.Census.ID.Hex(), processID)
+	if err != nil {
+		ErrGenericInternalServerError.WithErr(err).Write(w)
+		return
+	}
+	if pubCensus.Census.Type == db.CensusTypeSMSorMail ||
```
p4u, line 82: Use switch case here. For clarity of the code.

api/census.go
```diff
@@ -156,14 +157,15 @@ func (a *API) publishCensusHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	rootBytes := a.twofactor.Signer.BlindPubKey().Bytes()
```
lucasmenendez: Probably this logic could be in the `twofactor` package, for example:
```go
func (sk *SaltedKey) BlindPubKeyBytes() []bytes {
	return sk.BlindPubKey().Bytes()
}

func (sk *SaltedKey) BlindPubKeyHex() string {
	return hex.EncodeToString(sk.BlindPubKeyBytes())
}
```

twofactor/twofactor.go
```diff
@@ -336,17 +348,29 @@ func (tf *Twofactor) Auth(electionID []byte, authToken *uuid.UUID, authData []st
 		log.Warnf("verify challenge %s failed: %v", solution, err)
 		return AuthResponse{Error: "challenge not completed:" + err.Error()}
 	}
+	r, err := tf.NewBlindRequestKey()
+	if err != nil {
+		return AuthResponse{Error: "error getting new blind token:" + err.Error()}
+	}
+	tokenR := r.BytesUncompressed()
```
lucasmenendez: I think that the `NewBlindRequestKey` method could return the `BytesUncompressed` result instead a `blind.Point`. The `blind.Point` is never used so we can assume that our token is already the uncompressed []bytes.

twofactor/twofactor.go
```diff
@@ -336,17 +348,29 @@ func (tf *Twofactor) Auth(electionID []byte, authToken *uuid.UUID, authData []st
...
 
 func (tf *Twofactor) Sign(electionID, payload, tokenR internal.HexBytes, address string) AuthResponse {
-	signature, err := tf.SignECDSA(tokenR, payload, electionID)
+	r, err := blindsecp256k1.NewPointFromBytesUncompressed(tokenR)
+	if err != nil {
+		return AuthResponse{
+			Success: false,
+			Error:   err.Error(),
+		}
+	}
```
lucasmenendez: Same here, I think that the bytes-to-point conversion can be included into the `SignBlind` method.

lucasmenendez (approved):

api/process_bundles.go
```diff
@@ -0,0 +1,465 @@
...
+// Requires Manager/Admin role for the organization that owns the bundle. Returns 200 on success.
+// The processes array in the request must not be empty.
+func (a *API) updateProcessBundleHandler(w http.ResponseWriter, r *http.Request) {
+	bundleIDStr := chi.URLParam(r, "bundleId")
+	if bundleIDStr == "" {
+		ErrMalformedURLParam.Withf("missing bundle ID").Write(w)
+		return
+	}
+
+	bundleID, err := primitive.ObjectIDFromHex(bundleIDStr)
```
lucasmenendez, line 148: Nit: I think that this conversion should be placed in `db` package to prevent database logic in the API layer.

api/process_bundles.go
```diff
@@ -0,0 +1,465 @@
...
+	if err != nil {
+		ErrMalformedURLParam.Withf("invalid bundle ID").Write(w)
+		return
+	}
+
+	bundle, err := a.db.ProcessBundle(bundleID)
+	if err != nil {
+		ErrGenericInternalServerError.WithErr(err).Write(w)
+		return
+	}
```
lucasmenendez, line 257: Same here, I think that is better to convert the string to `ObjectID` into the `ProcessBundle` method.

api/process_bundles.go
```diff
@@ -0,0 +1,465 @@
...
+		} else if req.Phone != "" {
+			if !bytes.Equal(internal.HashOrgData(census.OrgAddress, req.Phone), participant.HashedPhone) {
+				return nil, ErrUnauthorized.Withf("invalid user data")
+			}
+			authResp = a.twofactor.InitiateAuth(bundleId, userID, req.Phone, notifications.SMS)
+		} else {
+			return nil, ErrUnauthorized.Withf("missing email or phone")
+		}
+	default:
+		return nil, ErrUnauthorized.Withf("invalid census type")
```
lucasmenendez, line 455: panic here (this code is unreachable) or remove the previous check (line 400):

```go
if censusType != db.CensusTypeMail && censusType != db.CensusTypeSMSorMail && censusType != db.CensusTypeSMS {
	return nil, ErrInvalidOrganizationData.Withf("invalid census type")
}
```

db/helpers.go
```diff
@@ -202,15 +207,34 @@ func (ms *MongoStorage) createIndexes() error {
...
 		Keys: bson.D{
 			{Key: "orgAddress", Value: 1},    // 1 for ascending order
 			{Key: "participantNo", Value: 1}, // 1 for ascending order
 		},
 		Options: options.Index().SetUnique(true),
 	}
-	if _, err := ms.orgParticipants.Indexes().CreateOne(ctx, orgParticipantNoIndex); err != nil {
+	if _, err := ms.orgParticipants.Indexes().CreateOne(ctx, orgParticipantdAddresNoIndex); err != nil {
 		return fmt.Errorf("failed to create index on orgAddress and participantNo for orgParticipants: %w", err)
 	}
```
lucasmenendez, line 239: You can use `CreateMany` with a `[]mongo.IndexModel` instead of call `CreateOne` repeatedly.

db/helpers.go
```diff
@@ -228,14 +252,34 @@ func (ms *MongoStorage) createIndexes() error {
...
 		Keys: bson.D{
 			{Key: "censusId", Value: 1},      // 1 for ascending order
 			{Key: "participantNo", Value: 1}, // 1 for ascending order
 		},
 		Options: options.Index().SetUnique(true),
 	}
-	if _, err := ms.censusMemberships.Indexes().CreateOne(ctx, censusMembershipIndex); err != nil {
+	if _, err := ms.censusMemberships.Indexes().CreateOne(ctx, censusMembershipIdNoIndex); err != nil {
 		return fmt.Errorf("failed to create index on censusID for censusMemberships: %w", err)
 	}
```
lucasmenendez, line 284: Same here

db/org_participants.go
```diff
@@ -207,3 +208,83 @@ func (ms *MongoStorage) BulkUpsertOrgParticipants(
...
+		}
+	}()
+
+	var orgParticipants []OrgParticipant
+	if err = cursor.All(ctx, &orgParticipants); err != nil {
+		return nil, fmt.Errorf("failed to get orgParticipants: %w", err)
+	}
+
+	return orgParticipants, nil
+}
```
lucasmenendez, line 239: I have almost the same method in my PR but paginated, so we must check how both can be merged to reuse code if is possible.

db/org_participants.go
```diff
@@ -207,3 +208,83 @@ func (ms *MongoStorage) BulkUpsertOrgParticipants(
...
+	}()
+
+	// Convert cursor to slice of OrgParticipants
+	var participants []CensusMembershipParticipant
+	if err := cursor.All(ctx, &participants); err != nil {
+		return nil, fmt.Errorf("failed to get orgParticipants: %w", err)
+	}
+
+	return participants, nil
+}
```
lucasmenendez, line 290: Same here, probably this can help to me to improve my `PaginatedCensusParticipants` method. Well done!

db/process_bundles.go
```diff
@@ -0,0 +1,265 @@
...
+	} else {
+		filter := bson.M{"_id": bundle.ID}
+		update := bson.M{"$set": bundle}
+		options := &options.UpdateOptions{}
+		options.SetUpsert(true)
+
+		if _, err := ms.processBundles.UpdateOne(ctx, filter, update, options); err != nil {
+			return "", fmt.Errorf("failed to update process bundle: %w", err)
+		}
+	}
```
lucasmenendez, line 59: Is this check required? Doesn't the upsert option do just that?

twofactor/types.go
```diff
@@ -0,0 +1,162 @@
...
+	// SignatureTypeBlind is a secp256k1 blind signature
+	SignatureTypeBlind = "blind"
+	// SignatureTypeEthereum is the standard secp256k1 signature used in Ethereum
+	SignatureTypeEthereum = "ecdsa"
+	// SignatureTypeSharedKey identifier the shared key (common for all users on the same processId)
+	SignatureTypeSharedKey = "sharedkey"
+)
+
+// AllSignatures is a helper list that includes all available CSP signature schemes.
+var AllSignatures = []string{SignatureTypeBlind, SignatureTypeEthereum, SignatureTypeSharedKey}
```
lucasmenendez, line 162: Nit: Move constants to the top.

lucasmenendez (changes requested):

PR #46, opened 2025-02-26
feature: new api structs

lucasmenendez:
Includes:
* New structs for the API to replace database structs and avoid encode/decode issues.
* New endpoint to get organization censuses.


db/census_membership.go
```diff
@@ -154,9 +154,8 @@ func (ms *MongoStorage) SetBulkCensusMembership(
 		return nil, fmt.Errorf("failed to get published census: %w", err)
 	}
 
-	org := ms.organizations.FindOne(ctx, bson.M{"_id": census.OrgAddress})
-	if org.Err() != nil {
-		return nil, ErrNotFound
+	if _, _, err := ms.Organization(census.OrgAddress, false); err != nil {
```
emmdim, line 157: This will throw because of the double lock that we discussed earlier

emmdim: Use the `ms.organization` instead

api/docs.md
```diff
@@ -44,7 +45,7 @@
   - [ 📄 Get object](#-get-object)
 - [📊 Census](#-census)
   - [📝 Create Census](#-create-census)
-  - [ℹ️ Get Census Info](#-get-census-info)
+  - [ℹ️ Get Census Info](#ℹ%EF%B8%8F-get-census-info)
```
emmdim, line 48: The extra characters should be removed. I don't thionk this hyperlink works

api/census.go
```diff
@@ -102,22 +90,31 @@ func (a *API) addParticipantsHandler(w http.ResponseWriter, r *http.Request) {
...
 	if !user.HasRoleFor(census.OrgAddress, db.ManagerRole) && !user.HasRoleFor(census.OrgAddress, db.AdminRole) {
 		ErrUnauthorized.Withf("user is not admin of organization").Write(w)
 		return
 	}
-
-	// add the org participants where necessary
-	no, err := a.db.SetBulkCensusMembership(passwordSalt, censusID, participantsInfo.Participants)
+	// decode the participants from the request body
+	participants := &AddParticipantsRequest{}
+	if err := json.NewDecoder(r.Body).Decode(participants); err != nil {
```
emmdim, line 100: Why do you do the decoding after check if the user has permissions?
Firsxt the requiest validity should be checked

db/census_membership.go
lucasmenendez: but there is no lock before this line, in fact we call `ms.Census` just above

emmdim (changes requested):
Please solve the comments

db/census_membership.go
emmdim: oooops :-|

api/census.go
lucasmenendez: Now, if the user does not have permission, the code stops and returns the error. With the previous code, the decoding is done regardless of whether the user meets the role requirements or not.

emmdim: So if there is a problematic argument why not avoid the user check?
In general i do no think what you propose is a good practice. Though considering the possible size of this request I think it can make sense

emmdim (approved):

PR #58, opened 2025-03-10
feature: csp refactor

lucasmenendez:
- [x] New package `csp` with more modular scheme.
- [x] Unused code removed.
- [x] Notification queue and notification challenges logic separated between them and from CSP logic
- [ ] CSP flow (handlers and responses) separated from internal implementation (blind, salted, ...)
- [x] Unit tests


csp/storage/storage.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/vocdoni/saas-backend/internal"
+)
+
+// Storage interface implements the storage layer for the smshandler
+type Storage interface {
```
p4u, line 11: Do we really need this interface?

p4u:
This the branch where I started to refactor. Maybe you can use it somehow (or maybe not). But at least take a quick look to the twofactor package.

https://github.com/vocdoni/saas-backend/tree/f/refactor/twofactor

csp/storage/storage.go
lucasmenendez: I don't know, right now i use it to guide the development.

csp/errors.go
```diff
@@ -0,0 +1,41 @@
+package csp
```
p4u, line 1: Consider using the errors/ package from the repository to define the errors

p4u (commented):
in general LGTM

PR #59, opened 2025-03-10
F/api tests

p4u:


subscriptions/subscriptions.go
```diff
@@ -91,6 +98,15 @@ func (p *Subscriptions) HasTxPermission(
 	org *db.Organization,
 	user *db.User,
 ) (bool, error) {
+	if org == nil {
+		return false, fmt.Errorf("organization is nil")
+	}
+
+	// Check if the organization has a subscription
+	if org.Subscription.PlanID == 0 {
+		return false, fmt.Errorf("organization has no subscription plan")
```
emmdim, line 107: PlanID 0 is a valid plan id. The check with 0 here does not bring the result you expected

subscriptions/subscriptions.go
```diff
@@ -141,6 +157,12 @@ func (p *Subscriptions) HasDBPersmission(userEmail, orgAddress string, permissio
 	if err != nil {
 		return false, fmt.Errorf("could not get organization: %v", err)
 	}
+
+	// Check if the organization has a subscription
+	if org.Subscription.PlanID == 0 {
```
emmdim, line 162: Same as above

PR #61, opened 2025-03-12
feature: csp sign flow

lucasmenendez:


csp/handlers/handlers.go
```diff
@@ -0,0 +1,251 @@
...
+				return nil, errors.ErrUnauthorized.Withf("invalid user phone")
+			}
+			toDestinations = req.Phone
+			challengeType = notifications.SMSChallenge
+		}
+	default:
+		return nil, errors.ErrNotSupported.Withf("invalid census type")
+	}
+	// parse the user ID and generate the token
+	userID := new(internal.HexBytes).SetString(participant.ID.Hex())
```
emmdim, line 229: @lucasmenendez this is incosistent with the way the userID is generated when creating the process:
https://github.com/vocdoni/saas-backend/blob/8c887c0f6a82971cd7d4eed5a184034e24732790/api/process_bundles.go#L132

PR #79, opened 2025-04-03
update swagger.yml (make swagger)

altergui:


emmdim:
@altergui i inderstand that i mixe the order of the PRs. I merged the stripe one first that includes this. IS that the case?

PR #80, opened 2025-04-03
feat(api): implements oauth login and register

emmdim:
More info in https://github.com/vocdoni/ui-scaffold/issues/864

### DIFF
--- a/api/auth.go
+++ b/api/auth.go
@@ -2,12 +2,15 @@ package api
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 
+	"github.com/vocdoni/saas-backend/account"
 	"github.com/vocdoni/saas-backend/api/apicommon"
 	"github.com/vocdoni/saas-backend/db"
 	"github.com/vocdoni/saas-backend/errors"
 	"github.com/vocdoni/saas-backend/internal"
+	"go.vocdoni.io/dvote/log"
 )
 
 // refreshTokenHandler godoc
@@ -99,6 +102,9 @@ func (a *API) authLoginHandler(w http.ResponseWriter, r *http.Request) {
 //	@Failure		401	{object}	errors.Error
 //	@Failure		404	{object}	errors.Error	"No organizations found"
 //	@Router			/auth/addresses [get]
+//
+// writableOrganizationAddressesHandler returns the list of addresses of the
+// organizations where the user has write access.
 func (a *API) writableOrganizationAddressesHandler(w http.ResponseWriter, r *http.Request) {
 	// get the user from the request context
 	user, ok := apicommon.UserFromContext(r.Context())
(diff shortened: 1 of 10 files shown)


altergui:
do `make swagger` and commit those changes in the PR as well :pray:

api/auth.go
```diff
@@ -126,3 +132,92 @@ func (a *API) writableOrganizationAddressesHandler(w http.ResponseWriter, r *htt
...
+			Email:     loginInfo.Email,
+			FirstName: loginInfo.FirstName,
+			LastName:  loginInfo.LastName,
+			Password:  internal.HexHashPassword(passwordSalt, loginInfo.UserOauthSignature),
+			Verified:  true,
+		}
+		if _, err := a.db.SetUser(user); err != nil {
+			errors.ErrGenericInternalServerError.WithErr(err).Write(w)
+		}
+	} else {
```
altergui, line 207: no need for else clause, you can close the if here and unindent the next 6 lines.

another option if you want to handle more cases in the future, is replacing the "if, if, else" with a switch
```go
switch user, err := a.db.UserByEmail(loginInfo.Email); {
case err == db.ErrNotFound:
		// Register the user
		// extract from the external signature the user pubkey and verify matches the provided one
case err != nil:
	errors.ErrGenericInternalServerError.Write(w)
	return
default:
		// Login
		// check that the address generated password matches the one in the database
}
```

api/routes.go
```diff
@@ -11,6 +11,8 @@ const (
 	authRefresTokenEndpoint = "/auth/refresh"
 	// POST /auth/login to login and get a JWT token
 	authLoginEndpoint = "/auth/login"
+	// POST /oauth/login to login verifying Oauth parameters Sand get a JWT token
```
altergui, line 14: i guess `s/Sand/and/`

altergui (approved):
i'd do `sed -i s/Oauth/OAuth/g *` everywhere. 

with that said, LGTM :+1:

api/auth.go
emmdim: Noted, no other cases are foreseen currently

errors/errors_definition.go
```diff
@@ -53,15 +53,15 @@ var (
...
+	ErrOrganizationNotFound       = Error{Code: 40049, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("organization not found")}
+	ErrNoOrganizations            = Error{Code: 40042, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("user has no organizations")}
+	ErrUserNotFound               = Error{Code: 40048, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("user not found")}
+	ErrNoOrganizationSubscription = Error{Code: 40040, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("organization has no subscription")}
+	ErrPlanNotFound               = Error{Code: 40043, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("subscription plan not found")}
+	ErrJobNotFound                = Error{Code: 40046, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("job not found")}
+	ErrCensusNotFound             = Error{Code: 40047, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("census not found")}
+	ErrCensusTypeNotFound         = Error{Code: 40048, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("census type not found")}
+	ErrCensusParticipantNotFound  = Error{Code: 40049, HTTPstatus: http.StatusNotFound, Err: fmt.Errorf("census participant not found")}
```
altergui, line 64: is changing error codes intended?

altergui (commented):
there are still lots of "Oauth"

i just did a `sed -i s/Oauth/OAuth/g */*.go */*/*.go` and it fixes everything, you can amend with that

errors/errors_definition.go
altergui: ok this was indeed a fix but unrelated to this PR, you'll commit this separate

PR #84, opened 2025-04-09
api: support creating (and listing) draft processes

altergui:
* [x] Rebase on top of current `main`
* [x] Add a new endpoint to list drafts (paginated) @altergui 
* [x] Rebase on top of #301 
* [x] remove the dependency on publishedCensusRoot
* [x] Take into acocunt that there are cases that the census can be null when storing a draft process
* [x] Add a PUT /proces/:id endpoint

### DIFF
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -255,6 +255,15 @@ type ListOrganizationMemberGroupResponse struct {
 	Members []OrgMember `json:"members"`
 }
 
+// ListOrganizationProcesses represents the response for listing the processes of an organization.
+// swagger:model ListOrganizationProcesses
+type ListOrganizationProcesses struct {
+	// Pagination fields
+	Pagination *Pagination `json:"pagination"`
+	// List of organization processes
+	Processes []db.Process `json:"processes"`
+}
+
 // ValidateMemberGroupRequest validates the request for creating or updating an organization member group.
 // Validates that either AuthFields or TwoFaFields are provided and checks for duplicates or empty fields.
 // swagger:model ValidateMemberGroupRequest
@@ -959,6 +968,20 @@ type CreateProcessRequest struct {
 	// Organization address
 	OrgAddress common.Address `json:"orgAddress"`
 
+	// Vochain ID/Address of the process
+	Address internal.HexBytes `json:"address" swaggertype:"string" format:"hex" example:"deadbeef"`
+
+	// Census ID
+	CensusID internal.HexBytes `json:"censusId" swaggertype:"string" format:"hex" example:"deadbeef"`
+
+	// Additional metadata for the process
+	// Can be any key-value pairs
+	Metadata map[string]any `json:"metadata"`
+}
+
+// UpdateProcessRequest defines the payload for updating an existing voting process.
+// swagger:model UpdateProcessRequest
+type UpdateProcessRequest struct {
 	// Vochain ID/Address of the process
 	Address internal.HexBytes `json:"address" swaggertype:"string" format:"hex" example:"deadbeef"`
 
--- a/api/process.go
+++ b/api/process.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net/http"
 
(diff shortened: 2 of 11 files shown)


altergui:
ready for review

emmdim (commented):
I see that the approach taken is to allow the db `Process` to be a draft.
This creates a problem with how processes were being thought so far in the backend.

The ID of the processes is thought to be the process address
https://github.com/vocdoni/saas-backend/blob/a02d97573ce25890f2c3178fd7ed7c4ecb9aa234/db/types.go#L367

Therefore it is assumed that this is a created process.

One solution to this that comes to mind is to follow the approach we follow with the other Mongo collections and make as unique identifier of the object a mongo `ObjectID`

api/process.go
emmdim: Take into acocunt that there are cases that the census can be null, according to how the UI behaves.
Therefore any use of the census obligatory should be removed 
This also affects the extraction of orgAdress from the census. `census.OrgAddress` which should be remove in favour of the orgAdress of  the request session

altergui:
great review!
i updated the base branch, 
i'll address the census issue you mentioned, and fix the rebase conflicts

api/apicommon/types.go
```diff
@@ -953,6 +964,26 @@ type CreateProcessRequest struct {
 	// Additional metadata for the process
 	// Can be any key-value pairs
 	Metadata map[string]any `json:"metadata"`
+
+	// Whether this process is a draft (not yet pushed to the vochain)
```
altergui: i'd append
```
// When draft=true, the process can be updated (overwritten) with PUT
```

api/process.go
```diff
@@ -82,6 +95,98 @@ func (a *API) createProcessHandler(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+			errors.ErrGenericInternalServerError.WithErr(err).Write(w)
+			return
+		}
+	}
+
+	if processInfo.Draft != nil {
+		existingProcess.Draft = *processInfo.Draft
+	}
```
altergui: existingProcess.Draft is not used at all, i.e. not checked before allowing update, so a user can update a process with PUT even after pushing to Vochain, AFAICS?

api/process_test.go
```diff
@@ -131,3 +144,96 @@ func TestProcess(t *testing.T) {
...
+	resp, code = testRequest(t, http.MethodGet, adminToken, nil, "process", pid)
+	c.Assert(code, qt.Equals, http.StatusOK, qt.Commentf("response: %s", resp))
+
+	var updatedProcess db.Process
+	err = parseJSON(resp, &updatedProcess)
+	c.Assert(err, qt.IsNil)
+	c.Assert(updatedProcess.Draft, qt.IsFalse, qt.Commentf("Process should no longer be in draft mode"))
+	c.Assert(updatedProcess.Metadata, qt.DeepEquals, updatedMetadata, qt.Commentf("Process metadata should be updated"))
+	t.Log("Verified process is no longer in draft mode and metadata was updated")
+
```
altergui: i had a test here `// Step 3: Try to update the process again, which should fail since it's no longer in draft mode` and it's gone, so is this design change on purpose?

api/process.go
```diff
@@ -82,6 +95,98 @@ func (a *API) createProcessHandler(w http.ResponseWriter, r *http.Request) {
...
+		existingProcess.Census = *census
+	}
+
+	_, err = a.db.SetProcess(existingProcess)
+	if err != nil {
+		errors.ErrGenericInternalServerError.WithErr(err).Write(w)
+		return
+	}
+
+	apicommon.HTTPWriteJSON(w, "Process updated successfully")
```
altergui: this looks odd in comparison to other endpoints, maybe just apicommon.HTTPWriteOK?

api/process_test.go
altergui: i reintroduced the test

altergui:
i'll rebase on top of current main to use the new pagination methods

Issue #69, opened 2025-03-27
db: refactor lock


altergui:
the current db code uses a Lock that is in many operations unneeded (since MongoDB has it's own lock). Some operations do require it (since MongoDB is not atomic) but many don't.
also the new implementation should likely be a RWLock (and not a simple Lock)

altergui:
![Image](https://github.com/user-attachments/assets/68ef5a2b-e263-449b-9e9c-cb3d95e10490)
![Image](https://github.com/user-attachments/assets/3d283130-d6d2-4502-99d9-a483e39a5ea1)

altergui:
i merged #87 and the remaining Lock AFAIU really just protects the Import and String methods, that need to be atomic (writes to db need to wait until String method finished, otherwise the resulting string will be inconsistent)

i believe further improvements could be achieved, but i'm closing this for now

PR #87, opened 2025-04-10
db: use RLock in readonly ops

altergui:
this is a small improvement addressing #69 

not yet tested deeply


altergui:
i removed all pointless locks (the ones protecting a single Find)
and converted some Lock -> RLock

it looks to me safe enough to merge, but testing this in "production" somehow would be ideal.

emmdim (approved):
Nice catch. Loogs good to me!

PR #95, opened 2025-04-29
csp: refactor to use db rather than mongodb

altergui:
- **csp: fix comments**
- **csp: refactor to use db rather than mongodb**


altergui:
tests are failing due to buggy code in my PR, i need to review all the code in `db/csp.go`

altergui:
i've double-checked every diff, and now it's looking good, ready for review and merge

db/errors.go
```diff
@@ -7,3 +7,22 @@ var (
 	ErrInvalidData   = fmt.Errorf("invalid data provided")
 	ErrAlreadyExists = fmt.Errorf("already exists")
 )
+
+var (
```
emmdim, line 11: What is the reason for creating a new var block here?

db/helpers.go
```diff
@@ -122,6 +122,16 @@ func (ms *MongoStorage) initCollections(database string) error {
...
+
+	// cspTokens collection
+	if ms.cspTokens, err = getCollection("cspTokens"); err != nil {
+		return err
+	}
+
+	// cspTokensStatus collection
+	if ms.cspTokensStatus, err = getCollection("cspTokensStatus"); err != nil {
+		return err
+	}
```
emmdim, line 134: Apart from initializing collections this is the file to create the corrsponing indexes.
Is there a cncrete reason why you did bring here the csp index:
https://github.com/vocdoni/saas-backend/blob/621acaf67f63e4366dee9e1e6413fdb712c7b472/csp/storage/mongodb.go#L139

csp/storage/csp_test.go
emmdim, line 1: Is thre a reason these tests are not moved to the merged db?

db/helpers.go
altergui: good catch! :raised_hands:

csp/storage/csp_test.go
altergui: good catch! recovered

db/errors.go
altergui: no reason at all, fixed

db/helpers.go
altergui: i re-added the index in  73e2dd6dad28698b695b538321a7c19566c33aaa

emmdim (approved):

Issue #105, opened 2025-05-08
Create endpoint for support mail response and forward


emmdim:
Implemenent a support endpoint for organizations that would receive the following information
```
Type of ticket: (The user selects it in the dropdown menu)
Name of the user:
Email of the user:
Organization ID: (Ethereum address of the organization)
Content:
```
and send it to our support email.

### Implementation

The new endpoint work only for authenticated users, and thus the user name and email are extracted from the DB.

```
POST /organizations/{orgID}/support
{
   type: "blah"
   title: ""
   body: "......"
}
```

The email sent to our support mailbox should have set as `ReplyTo` the address of the user

Relevant to:
https://github.com/vocdoni/ui-scaffold/issues/1035

cc @elboletaire

PR #112, opened 2025-05-15
Adding email support endpoint

emmdim:
Relevant to https://github.com/vocdoni/saas-backend/issues/105

### DIFF
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -853,3 +853,14 @@ type OAuthServiceAddressResponse struct {
 	// The address of the OAuth service signer
 	Address string `json:"address"`
 }
+
+type CreateOrganizationTicketRequest struct {
+	// Type of the ticket to create (definded externally)
+	TicketType string `json:"type"`
+
+	// Title of the ticket
+	Title string `json:"title"`
+
+	// Body of the ticket
+	Description string `json:"description"`
+}
--- a/api/docs.md
+++ b/api/docs.md
@@ -37,6 +37,7 @@
   - [🔍 Check Add Participants Job Status](#-check-add-participants-job-status)
   - [❌ Delete Organization Participants](#-delete-organization-participants)
   - [📋 Organization Meta Information](#-organization-meta-information)
+  - [🎫 Create Organization Ticket](#-create-organization-ticket)
   - [🤠 Available organization members roles](#-available-organization-members-roles)
   - [🏛️ Available organization types](#-available-organization-types)
 - [🏦 Plans](#-plans)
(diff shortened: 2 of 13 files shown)


emmdim:
ToD:
* [ ]  HTML template
* [ ] Apply template exec for Subject

api/apicommon/types.go
```diff
@@ -853,3 +853,14 @@ type OAuthServiceAddressResponse struct {
 	// The address of the OAuth service signer
 	Address string `json:"address"`
 }
+
+type CreateOrganizationTicketRequest struct {
+	// Type of te ticket to create (definde externally)
```
altergui, line 858: the 
defined

api/docs.md
```diff
@@ -1046,6 +1047,60 @@ Deletes multiple participants from an organization by their participant IDs. Req
...
+
+* **Errors**
+
+| HTTP Status | Error code | Message |
+|:---:|:---:|:---|
+| `401` | `40001` | `user not authorized` |
+| `400` | `40004` | `malformed JSON body` |
+| `400` | `40011` | `no organization provided` |
+| `500` | `50002` | `internal server error` |
+
```
altergui, line 1103: why twice? isn't this is exactly the same as lines 1050-1076?

api/routes.go
```diff
@@ -72,6 +72,8 @@ const (
 	organizationDeleteParticipantsEndpoint = "/organizations/{address}/participants"
 	// POST/PUT/GET/DELETE /organizations/{address}/meta to add/set/get/delete the organization metadata
 	organizationMetaEndpoint = "/organizations/{address}/meta"
+	// POST /organizations/{address}/ticket to create a new ticket to our support system
+	organizationCreateTicketEndpoint = "/organizations/{address}/tickets"
```
altergui, line 76: `POST /organizations/{address}/ticket` (no plural) sounds more reasonable to me (unless there are plans to implement a `GET /organizations/{address}/tickets` but i don't think that's the case, right?

altergui: in fact the comment on top says `ticket`

api/docs.md
emmdim: Oops. This was an error introduced when rebasing with main.

api/organizations.go
```diff
@@ -711,3 +711,79 @@ func (a *API) organizationCensusesHandler(w http.ResponseWriter, r *http.Request
...
+//	@Produce		json
+//	@Security		BearerAuth
+//	@Param			address	path		string										true	"Organization address"
+//	@Param			request	body		apicommon.CreateOrganizationTicketRequest	true	"Ticket request information"
+//	@Success		200		{string}	string										"OK"
+//	@Failure		400		{object}	errors.Error								"Invalid input data"
+//	@Failure		401		{object}	errors.Error								"Unauthorized"
+//	@Failure		404		{object}	errors.Error								"Organization not found"
+//	@Failure		500		{object}	errors.Error								"Internal server error"
+//	@Router			/organizations/{address}/tickets [post]
```
altergui, line 730: missed this one, should be ticket

altergui (approved):
LGTM, missing just one fix (and regenerate swagger.yml) but i'm approving already

PR #132, opened 2025-05-27
Make Invitatation management restful and add Resend Invite endpoint

emmdim:
Closes https://github.com/vocdoni/saas-backend/issues/126


altergui (approved):
LGTM

PR #124, opened 2025-05-26
all: rename concept org Member -> User

altergui:
progress towards #116 

Before merging:
[ ] Update stripe
[ ] Test with UI

PR #134, opened 2025-05-29
all: rename concept org Participant -> Member

altergui:
to be applied on top of #124
(in a separate PR, to ease reviewing the changes)

also includes rename MemberNo -> MemberID
and an odd MembersNo (actually a count) that was renamed to Count


emmdim:
MErged to https://github.com/vocdoni/saas-backend/pull/139

PR #155, opened 2025-06-13
MAke member id not unique and member importing based on internal IDs

### LINKED ISSUE #148: db: Make member internal ID as the only forced UID
* [x] Make ID as the UID in importing etc
* [x] Create not unique index on memberID org
* [x] Make indexes on phone and email not unique
* [x] Remove memberID as UID in importing and other 
* [ ] Remove old mongo indexes before merging
* [ ] Convert MemberID to memberNumber

emmdim:
Closes #148

### DIFF
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -9,6 +9,7 @@ import (
 	"github.com/vocdoni/saas-backend/db"
 	"github.com/vocdoni/saas-backend/internal"
 	"github.com/vocdoni/saas-backend/notifications"
+	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.vocdoni.io/dvote/log"
 )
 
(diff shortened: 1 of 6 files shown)


altergui (commented):
i'm lost on this one, to be honest, in terms of what is memberID and what is internalID. is there any alternative naming that would reduce confusion?

db/census_membership_test.go
```diff
@@ -26,11 +26,12 @@ func setupTestCensusMembershipPrerequisites(t *testing.T, memberSuffix string) (
 	// Create test member with unique ID
 	memberID := testMembershipMemberID + memberSuffix
```
altergui, line 27: should this var be renamed to memberNumber? for clarity / consistency

altergui (approved):
introducing the name "memberNumber" really improves clarity

Issue #156, opened 2025-06-13
Use internal MemberID for importing members throuh census creation


emmdim:
TEST with @jpaulet  before merging

Follow up from #148 

If Census Members importing is going to be used then an approach similar to #155 should be implemented for Census Memership as well.
If this way of importing members is not going to be used then it should be removed.

This should be done after any projects using the existing system.

PR #160, opened 2025-06-19
F/148 bis rename cenus membership

emmdim:
Follow-up of #155,  in separate PRs for easier review.
To merge only the current PR, after pointing it correctly to main

Closes https://github.com/vocdoni/saas-backend/issues/148
Closes https://github.com/vocdoni/saas-backend/issues/156

### DIFF
--- a/api/csp_voting_test.go
+++ b/api/csp_voting_test.go
@@ -197,21 +197,21 @@ func TestCSPVoting(t *testing.T) {
 
 				// Test cases to try to break the authentication and voting mechanisms
 				t.Run("Authentication Attack Vectors", func(_ *testing.T) {
-					// Test case 1: Try to authenticate with invalid member ID
-					t.Run("Invalid Member ID", func(_ *testing.T) {
+					// Test case 1: Try to authenticate with invalid participant ID
+					t.Run("Invalid Participant ID", func(_ *testing.T) {
 						authReq := &handlers.AuthRequest{
-							MemberID: "INVALID",
-							Email:    "john.doe@example.com",
+							ParticipantID: "INVALID",
+							Email:         "john.doe@example.com",
 						}
 						resp, code := testRequest(t, http.MethodPost, "", authReq, "process", "bundle", bundleID, "auth", "0")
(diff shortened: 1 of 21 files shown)


api/csp_voting_test.go
```diff
@@ -200,8 +200,8 @@ func TestCSPVoting(t *testing.T) {
 					// Test case 1: Try to authenticate with invalid member ID
```
altergui, line 200: comment and test name should be adapted as well

api/csp_voting_test.go
```diff
@@ -210,8 +210,8 @@ func TestCSPVoting(t *testing.T) {
 					// Test case 2: Try to authenticate with valid member ID but wrong email
```
altergui, line 210: same here, likely in many more places

csp/handlers/handlers.go
```diff
@@ -361,8 +361,8 @@ func (c *CSPHandlers) ConsumedAddressHandler(w http.ResponseWriter, r *http.Requ
 }
 
 // validateMemberNumber checks if the member number is provided
-func validateMemberNumber(memberID string) error {
-	if len(memberID) == 0 {
+func validateMemberNumber(memberNumber string) error {
```
altergui, line 364: this should happen in the other PR. but maybe we can squash both together and name it "replace memberID with memberNumber and participantID"

db/census_participant.go
```diff
@@ -118,13 +118,13 @@ func (ms *MongoStorage) CensusMembership(censusID, id string) (*CensusMembership
 	return membership, nil
 }
 
-// CensusMembership retrieves a census membership from the database based on
+// CensusParticipant retrieves a census membership from the database based on
 // memberNumber and censusID. Returns ErrNotFound if the membership doesn't exist.
-func (ms *MongoStorage) CensusMembershipByMemberNumber(
+func (ms *MongoStorage) CensusParticipantByMemberNumber(
```
altergui, line 123: love the renaming, now this method is much more clear what is doing

altergui (commented):
didn't finish the review but loving it already, posting some comments

csp/handlers/handlers.go
altergui: will be separate PRs

PR #169, opened 2025-07-04
api: return errors on addMembers job

altergui:
api: return errors on addMembers job
    
    * db: merge processOrgMemberBatch into createOrgMemberBulkOperations
    * db: refactor startOrgMemberProgressReporter for simplicity
    * db: rename BulkOrgMembersStatus -> BulkOrgMembersJob
    * db: add Errors field to BulkOrgMembersJob


emmdim (approved):
Nice job. I liked the mini refactor approach

PR #183, opened 2025-07-10
    all: OrgAddress is now an internal.HexBytes

### LINKED ISSUE #189: api: review OrgAddress handling and make it consistent
first proposal https://github.com/vocdoni/saas-backend/pull/183

altergui:
internal: refactor HexBytes
    
    * make sanitize flow consistent
    * methods now use pointer receiver only when needed (so String method is called on fmt.Print for example)
    * new method MarshalBSONValue so it can be used in bson
    * new func HexBytesFromString

    all: OrgAddress is now an internal.HexBytes


emmdim:
Closed in favour of the:
https://github.com/vocdoni/saas-backend/pull/191

PR #192, opened 2025-07-16
api(test): dedup code with new funcs requestAndParse, requestAndAssertCode, etc

altergui:
this greatly simplifies the code, and always having `qt.Commentf(resp)` makes debugging regressions much easier


emmdim:
Did you skipp those 3 in api/api_test.go in purpose?

2025-07-16-172455_891x552_scrot

LGTM in general!

PR #198, opened 2025-07-18
all: removes censusType from CreateCensus requests. Updates setCensus…

emmdim:
… in db to extract censustype from TwoFaFields


altergui (approved):
LGTM

PR #200, opened 2025-07-24
db: new type Phone

altergui:


db/types.go
```diff
@@ -320,3 +324,173 @@ type ProcessesBundle struct {
...
+		return err
+	}
+
+	p.original = phoneStr
+	p.isHashed = false
+	return nil
+}
+
+// SetOrgAddress sets the organization address for hashing operations
+func (p *Phone) SetOrgAddress(orgAddress common.Address) {
```
emmdim, line 483: The naming of the function seems a bit weird since the result could also be the hashing of the data

db/org_members_groups.go
emmdim, line 1: does the `mapKeysToSlice` work correclty with the new Phone type?

emmdim (changes requested):
I really like the approach. Left some minor comments though

db/org_members_groups.go
altergui: good question. the answer is yes, but i extended the TestOrganizationMemberGroup to cover this and ensure no regressions.

PR #212, opened 2025-08-15
db: now SanitizeAndVerifyPhoneNumber uses Org country code when available

altergui:


db/org_members.go
```diff
@@ -23,52 +24,34 @@ func (ms *MongoStorage) SetOrgMember(salt string, orgMember *OrgMember) (string,
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
 	defer cancel()
 
-	if orgMember.OrgAddress.Cmp(common.Address{}) == 0 {
-		return "", ErrInvalidData
-	}
-
```
altergui, line 29: i removed this as redundant since the next check ms.Organization(orgMember.OrgAddress) will anyway return ErrInvalidData if OrgAddress is the zero address

db/types.go
```diff
@@ -206,7 +204,8 @@ type OrgMember struct {
 	// OrgAddress can be used for future sharding
 	OrgAddress     common.Address `json:"orgAddress" bson:"orgAddress"`
 	Email          string         `json:"email" bson:"email"`
-	Phone          *Phone         `json:"phone" bson:"phone"`
+	Phone          HashedPhone    `json:"phone" bson:"phone"`
```
emmdim, line 207: It seems a bit confusing in other parts of the code reading _Phone_ and meaning the _HashedPhone_, since in the API the text input is also named _Phone_

What do you think?

emmdim (approved):
Amazing. Great work from finding the issue till solving it and improving relevant parts on the way.

PR #214, opened 2025-08-17
f/use objectid everywhere

altergui:
to be discussed


altergui:
looks good enough and makes the original PR much better, will merge and discuss there

PR #242, opened 2025-09-15
simplify stripe package

altergui:


api/stripe_webhook_test.go
altergui: this whole file was fundamentally rewritten so the diff is impossible to follow

PR #266, opened 2025-09-25
db: fix ListOrganizationMemberGroup when pageSize==0

altergui:
the new test reproduces the original issue, here's the failure before the fix

image

### DIFF
--- a/db/org_members.go
+++ b/db/org_members.go
@@ -620,8 +620,10 @@ func (ms *MongoStorage) orgMembersByIDs(
 	if err != nil {
 		return 0, nil, err
 	}
-	totalPages := int(math.Ceil(float64(totalCount) / float64(pageSize)))
-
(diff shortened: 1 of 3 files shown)


db/org_members.go
```diff
@@ -620,8 +620,10 @@ func (ms *MongoStorage) orgMembersByIDs(
 	if err != nil {
 		return 0, nil, err
 	}
-	totalPages := int(math.Ceil(float64(totalCount) / float64(pageSize)))
-
+	totalPages := 1 // by default, with pageSize == 0, we fetch all members in one page
+	if pageSize > 0 {
+		totalPages = int(math.Ceil(float64(totalCount) / float64(pageSize)))
```
emmdim, line 625: If `pageSize > totalCount` then we still get `totalPages=0` which also leads to weird API results, right?

emmdim (approved):

PR #268, opened 2025-09-26
api: updatePendingUserInvitationHandler now checks invite belongs to org.Address

altergui:


altergui:
i went further and made OrganizationInvite.Role be a db.UserRole

PR #274, opened 2025-10-02
api: new middleware to set `lang` parameter

altergui:
- **api: new middleware to set lang parameter in context**

__New Middleware: `setLang`__

- __Location__: `api/middlewares.go`
- __Purpose__: Extracts language parameters from HTTP requests and stores them in the request context for downstream handlers to use

__Key Implementation Details:__

1. __Parameter Sources__: The middleware checks two sources for the language parameter:

   - __URL Parameters__: Uses `chi.URLParam(r, "lang")` to extract from route parameters (`/request/:lang`)
   - __Query Parameters__: Uses `r.URL.Query().Get("lang")` to extract from query string (`/request?lang=en`)

2. __Context Storage__: Both sources store the language value using the `apicommon.LangMetadataKey` constant as the context key

3. __Constant Definition__:

   - __Location__: `api/apicommon/const.go`
   - Added `LangMetadataKey MetadataKey = "lang"` constant

4. __Router Integration__:

   - __Location__: `api/api.go`
   - Middleware is applied globally using `r.Use(a.setLang)` in the router setup

__Functionality__: This middleware enables internationalization (i18n) support by making language preferences available to all handlers through the request context, allowing them to serve localized content based on user language preferences passed either as URL path parameters or query parameters.


altergui:
ready to review.
(i'm not 100% convinced about the namings in mailtemplates definitions.go and templates.go, but i'll leave it as-is for now)

altergui:
> ready to review. (i'm not 100% convinced about the namings in mailtemplates definitions.go and templates.go, but i'll leave it as-is for now)

i rewrote the namings in mailtemplates and i'm now satisfied, so this PR is now 100% ready for review+merge

emmdim (approved):
Nice. Very meticulous work

PR #280, opened 2025-10-07
db: DeleteAllOrgMembers now leaves empty groups

altergui:


altergui:
> The issues that this is just the unpatsed JSON response of the endpoint, and for some reason thre are two objects there that are not inside an array :-P

so, this bug is longstanding, unrelated to this PR, and I've fixed it in #293

so we can merge this one as-is and review #293 afterwards

PR #283, opened 2025-10-08
Enable Monthly and Yearly Stripe Pricing

### LINKED ISSUE #279: Adapt stripe service and handlers  to manage both monthly and annual subscriptions
### Description
In the current stripe implementation. each product is allowed to have only one price associated, using the default price for that. 
This plan was also asociated with the plans in the db:

https://github.com/vocdoni/saas-backend/blob/9ba8fc6b8251577cf872a0b6a14e840c3faa36da/db/types.go#L115-L119

Also the plans allowed so far where only yearly.

Therefore these two new changes need to be contemplated.

### Acceptance Criteria
* [ ] Stripe
   * [ ] Should trace automatically all the active prices, and separate the montly from the annual ones
   * [ ] in the checkout the client should be able  to define the billing period desired (yearly, annual)
   * [ ] Webhooks
      * [ ] Receive price updates 
      * [ ] identify correctly when a recurrent payment fails or not
  
* [ ] plans
   * [ ] HAndle priceIDs when necesarry ( either exted the plan to store also the pmontly price, or remove prices and search for them upon checkout everytime

emmdim:
Closes #279 

This refactor transforms the subscription system from single-price plans to support both monthly and yearly billing cycles.

### Database Schema Updates
- **Plan struct**: Replaced single `StripePriceID`/`StartingPrice` with separate monthly/yearly price fields (`StripeMonthlyPriceID`, `MonthlyPrice`, `StripeYearlyPriceID`, `YearlyPrice`)
- **OrganizationSubscription**: Added `BillingPeriod` and `StripeSubscirptionID` fields to track billing cycles
- **New BillingPeriod type**: Added constants for `month` and `year` billing periods

### Stripe Integration Enhancements
- **New GetProductPrices method**: Retrieves all active prices for a Stripe product
- **Enhanced checkout flow**: `CreateCheckoutSessionWithLookupKey` now accepts billing period parameter to select appropriate price ID
- **Improved subscription parsing**: Extracts billing period from Stripe webhook events
- **Product-to-plan processing**: Now handles multiple prices per product and validates both monthly/yearly prices exist

### API Updates
- **SubscriptionCheckout**: Added `billingPeriod` field for checkout requests
- **SubscriptionPlan**: Updated to expose both monthly and yearly pricing information
- **Enhanced error handling**: Validates billing period selection and ensures both price types are available

### Additional Improvements
- Added 15-day trial period to checkout sessions
- Updated all tests and mock data to use dual-pricing structure
- Improved webhook handling for product updates with multiple prices

This change maintains backward compatibility while enabling flexible billing options for customers.

### DIFF
--- a/stripe/service.go
+++ b/stripe/service.go
@@ -382,13 +382,14 @@ func processProductToPlan(planID uint64, product *stripeapi.Product, prices []st
 	}
 
 	plan := &db.Plan{
-		ID:           planID,
-		Name:         product.Name,
-		StripeID:     product.ID,
-		Default:      isDefaultPlan(product),
-		Organization: organizationData,
-		VotingTypes:  votingTypesData,
-		Features:     featuresData,
+		ID:            planID,
+		Name:          product.Name,
+		StripeID:      product.ID,
+		Default:       isDefaultPlan(product),
+		Organization:  organizationData,
+		VotingTypes:   votingTypesData,
+		Features:      featuresData,
+		FreeTrialDays: 0,
 	}
 
 	for _, price := range prices {
--- a/db/plans.go
+++ b/db/plans.go
@@ -47,7 +47,7 @@ func (ms *MongoStorage) SetPlan(plan *Plan) (uint64, error) {
 		}
 		return plan.ID, nil
 	}
-	updateDoc, err := dynamicUpdateDocument(plan, nil)
+	updateDoc, err := dynamicUpdateDocument(plan, []string{"freeTrialDays"})
 	if err != nil {
 		return 0, err
 	}


stripe/service.go
```diff
@@ -270,14 +289,27 @@ func (s *Service) CreateCheckoutSessionWithLookupKey(
 		return nil, NewStripeError("org_not_found", fmt.Sprintf("organization %s not found", orgAddress), err)
 	}
 
+	// Determine if the organization is eligible for a free trial based on
+	// if they already made a subscription and request a yearly plan
+	freeTrialEligible := org.Subscription.LastPaymentDate.IsZero() && billingPeriod == string(db.BillingPeriodAnnual)
```
altergui: why another name for the same concept? could be hasFreeTrial
but my suggestion would be to turn this into 
```go
if org.Subscription.LastPaymentDate.IsZero() && billingPeriod == string(db.BillingPeriodAnnual) {
  params.TrialPeriodDays = s.config.FreeTrialDays
}
```
after the `params := &CheckoutSessionParams`

stripe/service.go
```diff
@@ -340,17 +381,34 @@ func processProductToPlan(planID uint64, product *stripeapi.Product) (*db.Plan,
...
-	return &db.Plan{
+	plan := &db.Plan{
 		ID:            planID,
 		Name:          product.Name,
-		StartingPrice: product.DefaultPrice.UnitAmount,
 		StripeID:      product.ID,
-		StripePriceID: product.DefaultPrice.ID,
 		Default:       isDefaultPlan(product),
 		Organization:  organizationData,
+		FreeTrialDays: s.config.FreeTrialDays,
```
altergui: this is mixing sources of truth, why not define FreeTrialDays in stripe? 

then for example
```go
	freeTrialDays, err := extractPlanMetadata[int64](product.Metadata["freeTrialDays"])
	if err != nil {
		return nil, err
	}
```

db/types.go
```diff
@@ -113,24 +113,38 @@ type Features struct {
...
-	Features      Features    `json:"features" bson:"features"`
+	ID                   uint64      `json:"id" bson:"_id"`
+	Name                 string      `json:"name" bson:"name"`
+	StripeID             string      `json:"stripeID" bson:"stripeID"`
+	StripeMonthlyPriceID string      `json:"stripeMonthlyPriceID" bson:"stripeMonthlyPriceID"`
+	MonthlyPrice         int64       `json:"monthlyPrice" bson:"monthlyPrice"`
+	StripeYearlyPriceID  string      `json:"stripeYearlyPriceID" bson:"stripeYearlyPriceID"`
+	YearlyPrice          int64       `json:"yearlyPrice" bson:"yearlyPrice"`
+	Default              bool        `json:"default" bson:"default"`
+	FreeTrialDays        int         `json:"freeTrialDays" bson:"freeTrialDays"`
```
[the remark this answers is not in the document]
emmdim: Since it is a result of Atoi an conversion a cast should take place anyways

PR #290, opened 2025-10-16
api: limit creation of process drafts

altergui:
api: limit creation of process drafts
    
    * extend TestDraftProcess to check limits enforcement
    
    * cleanup leftovers introduced in "api: support creating a draft process"

### DIFF
--- a/api/process.go
+++ b/api/process.go
@@ -9,6 +9,7 @@ import (
 	"github.com/vocdoni/saas-backend/api/apicommon"
 	"github.com/vocdoni/saas-backend/db"
 	"github.com/vocdoni/saas-backend/errors"
+	"github.com/vocdoni/saas-backend/subscriptions"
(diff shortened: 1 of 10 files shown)


api/process.go
```diff
@@ -86,6 +86,38 @@ func (a *API) createProcessHandler(w http.ResponseWriter, r *http.Request) {
...
+			errors.ErrOrganizationNotFound.WithErr(err).Write(w)
+			return
+		}
+
+		if org.Subscription.PlanID == 0 {
+			errors.ErrNoOrganizationSubscription.With("can't create draft process").Write(w)
+			return
+		}
+
+		plan, err := a.db.Plan(org.Subscription.PlanID)
```
emmdim: Do you think it would make sense to have checks like these here:
https://github.com/vocdoni/saas-backend/blob/6f2569b756c7873c329e87c93edae20660d56bfb/subscriptions/subscriptions.go#L167

?

PR #296, opened 2025-10-20
fix api test anti-patterns

altergui:
- **add scripts/check-qt-patterns.sh**
- **api(test): use qt.HasLen rather than c.Assert(len(), qt.Equals)**
- **api(test): use qt.IsTrue rather than qt.Equals, true**


altergui:
this is just style changes in test files. no need to lose time reviewing carefully, honestly, can gloss over, approve & merge

PR #324, opened 2025-11-19
csp: During voter login checks census participants by 2fa fields only if 2fa fields are provided by the census

emmdim:


elboletaire (approved):
LGTM, we should remember to repeat that buggy login behavior with a 2fa process after this has been merged.

altergui (approved):
After rereading the code many times, i still don't understand if this fixes a bug (some vulnerability?) or it's just an optimization to skip a db query. 

in any case, LGTM so here's my approval as well

elboletaire:
> After rereading the code many times, i still don't understand if this fixes a bug (some vulnerability?) or it's just an optimization to skip a db query.
> 
> in any case, LGTM so here's my approval as well

There's some error related to how the errors are managed there. Without the  fix, if you try to login with incorrect data to a non 2FA  process, you'll see the following error:

image

Since this error is only intended for 2FA processes, as a workaround, we thought to limit this check only to 2FA processes. But there's some underlying bug as I said here, I said to give it a try with 2fa to see if the behavior with this fix is enough, or we should investigate it further to properly find the error flow issue.

altergui:
this is ready to merge, from my POV

PR #331, opened 2025-11-25
refactor(auth): replace hashing with encryption for verification codes

emmdim:
## Improved Verification Code Security with AES-GCM Encryption

This PR replaces the simple hash-based verification code storage with AES-GCM encryption, improving security while maintaining short verification codes in email links.

### Changes

**New encryption utilities** (`internal/utils.go`):
- Added `EncryptTokenToHex()`: Encrypts verification tokens using AES-GCM with argon2-derived keys
- Added `DecryptTokenFromHex()`: Decrypts stored tokens back to their original form
- Uses email as both key derivation salt and additional authenticated data (AAD) for binding

**Updated verification code handling** (`api/helpers.go`, `api/users.go`):
- `generateVerificationCodeAndLink()`: Now encrypts codes before database storage (previously used simple hashing)
- `verifyUserAccountHandler()`: Decrypts stored codes for verification
- `resendUserVerificationCodeHandler()`: Decrypts codes to regenerate verification links with original short codes
- `resetUserPasswordHandler()`: Decrypts password reset codes before user lookup

**Test coverage** (`internal/utils_test.go`, `api/users_test.go`):
- Comprehensive encryption/decryption tests covering valid cases, error conditions, and tampering detection
- Verified that different nonces produce different ciphertexts for same inputs
- Added verification link extraction and validation in user tests

### Benefits

- **Enhanced security**: Encrypted storage with authenticated encryption (GCM mode)
- **Maintained UX**: Short verification codes remain in email links (6 characters)
- **Email binding**: Tokens are cryptographically bound to user emails, preventing reuse across accounts
- **Backward compatibility**: Uses existing argon2 key derivation infrastructure

Closes https://github.com/vocdoni/saas-backend/issues/328

### DIFF
--- a/api/helpers.go
+++ b/api/helpers.go
@@ -127,10 +127,13 @@ func (a *API) generateVerificationCodeAndLink(target any, codeType db.CodeType) 
 			return "", "", fmt.Errorf("invalid target type")
 		}
 		// generate the verification code for the user and the expiration time
-		hashCode := internal.HashVerificationCode(user.Email, verificationCode)
+		sealedCode, err := internal.SealToken(verificationCode, user.Email, a.secret)
+		if err != nil {
+			return "", "", err
+		}
 		exp := time.Now().Add(apicommon.VerificationCodeExpiration)
 		// store the verification code in the database
-		if err := a.db.SetVerificationCode(&db.User{ID: user.ID}, hashCode, codeType, exp); err != nil {
+		if err := a.db.SetVerificationCode(&db.User{ID: user.ID}, sealedCode, codeType, exp); err != nil {
 			return "", "", err
 		}
 		// set the web app URI and the link parameters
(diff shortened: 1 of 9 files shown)


api/helpers.go
```diff
@@ -127,7 +127,11 @@ func (a *API) generateVerificationCodeAndLink(target any, codeType db.CodeType)
 			return "", "", fmt.Errorf("invalid target type")
 		}
 		// generate the verification code for the user and the expiration time
-		hashCode := internal.HashVerificationCode(user.Email, verificationCode)
+		// hashCode := internal.HashVerificationCode(user.Email, verificationCode)
```
altergui: just remove the line

altergui (commented):
in general LGTM but i find it confusing that both plaintext and ciphertext are the same type. i think it would be better to have different types ([]byte for ciphertext and string for plaintext). i'll try this out myself and see how it looks like

altergui (approved):

Issue #299, opened 2025-10-22
Allow more date formats during import of members


elboletaire:
We only allow a single date format when importing `YYYY-MM-DD`. This is OK for now, but we should add more possibilities here, since many clients probably won't download the template, or even if they download it, they probably won't know how to change this 🙄 

I propose adding three more parsing templates:

- ` yyyy/mm/dd`
- `dd/mm/yyyy`
- dd-mm-yyyy`

Because this will import data from users that might be exporting them into multiple formats (and without the possibility or know-how to alter these...), I recommend creating a generic time parse function that has **all types** into consideration, and use it where we're currently using `time.Parse`:

```go
package dateutil

import (
	"errors"
	"strings"
	"time"
)

// ParseDateAny tries many common date layouts using only the standard library.
// It prefers ISO (YYYY-MM-DD), then DMY (DD/MM/YYYY), and optionally MDY (U.S. style).
// - preferDMY controls what to do when a string could be either DMY or MDY and both parse:
//   if true, DMY is tried before MDY; if false, MDY is tried before DMY.
// - loc is used for ParseInLocation to avoid unexpected TZ shifts when no timezone is present.
//   Pass time.Local or a specific location; if nil, time.Local is used.
func ParseDateAny(s string, preferDMY bool, loc *time.Location) (time.Time, error) {
	if loc == nil {
		loc = time.Local
	}
	in := strings.TrimSpace(s)
	if in == "" {
		return time.Time{}, errors.New("empty date string")
	}

	// Normalize common separators that sometimes appear doubled or mixed.
	// We keep original too, some layouts below accept slashes vs dashes directly.
	norm := strings.NewReplacer(
		".", "-",
		"_", "-",
		"–", "-",
		"—", "-",
		"  ", " ",
	).Replace(in)
	norm = strings.TrimSpace(norm)

	// Base layouts (date-only). We also include a couple of tolerant variants
	// that allow single-digit month/day (Go requires exact layout, so we list a few).
	iso := []string{
		"2006-01-02",
		"2006/01/02",
		"2006-1-2",
		"2006/1/2",
	}
	dmy := []string{
		"02/01/2006",
		"02-01-2006",
		"2/1/2006",
		"2-1-2006",
		// User mentioned odd "DD-MM/YYYY" → include just in case:
		"02-01/2006",
		"2-1/2006",
	}
	mdy := []string{
		"01/02/2006",
		"01-02-2006",
		"1/2/2006",
		"1-2-2006",
[...]

PR #334, opened 2025-11-26
refactor(birth date): allow more formats

elboletaire:
The american format was considered as a final fallback, but would have been weird to accept it for some dates and not for others (since it was being added as a fallback of dd/mm/yyyy, some dates could land as dd/mm/yyyy when they were expected to be mm/dd/yyyy)

Also, you might be wondering why didn't I simply use `time.Parse()`, and that was to allow accepting multiple different separators. A straight time.Parse fallback (e.g., trying 02/01/2006 after 2006-01-02) would still need extra logic to reject ambiguity and handle mixed separators.

Closes https://github.com/vocdoni/saas-backend/issues/299

### DIFF
--- a/db/org_members.go
+++ b/db/org_members.go
@@ -183,14 +183,12 @@ func prepareOrgMember(org *Organization, m *OrgMember, salt string, currentTime 
 
 	// Check that the birthdate is valid
 	if len(member.BirthDate) > 0 {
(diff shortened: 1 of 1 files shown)


emmdim (approved):
Implementation looks good. Will move the definition of birthdate and its test to the internal modules

PR #342, opened 2025-11-30
fix mongodb timeouts during `go test`

altergui:
on one hand, i increased the ulimit so mongodb never runs out of resources.

on the other hand, i improved the Reset method so tests don't consume so much mongodb resources, and at run MUCH faster (20x)

### DIFF
--- a/api/transaction_test.go
+++ b/api/transaction_test.go
@@ -27,7 +27,7 @@ const (
 func TestSignTxHandler(t *testing.T) {
 	c := qt.New(t)
 	defer func() {
-		if err := testDB.Reset(); err != nil {
+		if err := testDB.DeleteAllDocuments(); err != nil {
 			c.Logf("error resetting test database: %v", err)
 		}
 	}()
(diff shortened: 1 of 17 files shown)


elboletaire:
> before [9d23253](https://github.com/vocdoni/saas-backend/commit/9d23253b43e9e47930b740a5f0a8526a3303660d)
> 
> ```
> ok  	github.com/vocdoni/saas-backend/db	173.959s
> ```
> 
> after
> 
> ```
> ok  	github.com/vocdoni/saas-backend/db	10.345s
> ```

Not only that.. now it even works from my computer 🙈

altergui:
> Nice effort, but weirdly enough i still get issues running the tests locally both in main and this branch. Especialy when running all the api tests together for example, when reaching the `api/lang_test.go` (after ~37sec) the tests fail due to a writeLoop @altergui

that sounds like a different bug, i can help debug it, but let's merge this PR, and open an issue for the writeLoop failure

emmdim (approved):
LGTM

PR #345, opened 2025-12-03
api: removes redundant birddate parsing during type conversion from api to db since this conversion takes place in the db

emmdim:
…


altergui (approved):
LGTM, just for reference this should have been included in cb079e175f (#299)

PR #390, opened 2026-01-07
feat(csp): count SentSMS and SentEmails

### LINKED ISSUE #373: Implement counting and limitation of emails and SMS per organization
* Add counting
* Enforce limiting in the right place https://github.com/vocdoni/saas-backend/blob/1869b501cfab1e8c070b2b000099c9304c6dc52e/subscriptions/subscriptions.go#L199

altergui:
- **api: remove OrganizationName (actually org.Address.Hex) from members_import_done email**
    csp: count SentSMS and SentEmails

Closes #373


api/apicommon/types.go
```diff
@@ -540,6 +540,12 @@ type SubscriptionPlanLimits struct {
 
 	// Whether this is a custom plan
 	CustomPlan bool `json:"customPlan"`
+
+	// Maximum number of sent emails allowed
+	MaxSentEmails int `json:"maxSentEmails" bson:"maxSentEmails"`
+
+	// Maximum number of sent SMS allowed
+	MaxSentSMS int `json:"maxSentSMS" bson:"maxSentSMS"`
```
altergui: this needs to be added on stripe side, right?

db/organizations.go
```diff
@@ -478,3 +478,70 @@ func (ms *MongoStorage) DecrementOrganizationSubOrgsCounter(address common.Addre
 
 	return nil
 }
+
+// IncrementOrganizationSentEmailsCounter atomically increments the emails counter for the organization with the given address.
+func (ms *MongoStorage) IncrementOrganizationSentEmailsCounter(address common.Address) error {
+	ms.keysLock.Lock()
```
emmdim: Considering that this is a write lock is in a part of the code that can have intensive load (voter login) we should ensure:
1) it does not block the rest of the backend usage (this is simple using a special lock)
2) ensure tha the rest of the login/vote process is not affected by this lock

What do you think @altergui

emmdim (changes requested):
Please address the comments

PR #409, opened 2026-01-29
fix(csp): remove cooldown time from auth-only token

### LINKED ISSUE #395: CSP auth cooldown should only be for 2fa authentications
We have a cooldown in the CSP auth method that should be only used for 2fa methods (email, sms), but right now is also being used for non 2fa methods, which is confusing and a bad UX.

**Steps to reproduce:**
- Create a process with a CSP census selecting NON 2fa fields, like member id
- Go to the process and login with a valid user
- Logout
- Login again, it will fail with the error "authentication required: attempt cooldown time not reached"

**Expected:**
To not have an enforced cooldown when there's no reason for (i.e. when not using 2fa fields)

altergui:
fix #395


emmdim (approved):
LGTM

PR #435, opened 2026-04-21
feat(groups): autogroup all members

elboletaire:
### Summary                                                                                                   
                                                                                                               
 When a user uploads a memberbase (or adds the first member), the platform now automatically creates an "All   
 members" group that always mirrors the entire memberbase. This eliminates the need to manually create a group 
 before starting a voting process, making the onboarding flow significantly smoother.                          
                                                                                                               
 ### How it works                                                                                              
                                                                                                               
 Rather than maintaining a real list of member IDs in sync, the auto group is stored as a lightweight document 
 with isAutoGroup: true and no stored member IDs. Its membership is resolved dynamically at read time by       
 querying the orgMembers collection directly — so it is always perfectly in sync with zero maintenance         
 overhead.                                                                                                     
                                                                                                               
 ### Behaviour                                                                                                 
                                                                                                               
 - Created automatically when the first member is added to an organization (via bulk upload, single upsert, or 
 CSV import)                                                                                                   
 - Always appears first in the groups listing (page 1 only, never duplicated on subsequent pages)              
 - Cannot be deleted — returns 403 Forbidden                                                                   
 - Membership cannot be manually modified — returns 403 Forbidden; title and description can still be updated  
 - Disappears automatically when all members are removed, and reappears when a member is added again           
 - Works transparently with the census flow (validation, creation) since the group has a real MongoDB ObjectID 
 - The isAutoGroup: true flag is exposed in API responses so frontends can display appropriate messaging       
                                                                                                               
 ### Changes                                                                                                   
                                                                                                               
 - db/types.go — added IsAutoGroup bool to OrganizationMemberGroup; added AutoGroupTitle /                     
 AutoGroupDescription constants                                                                                
 - db/errors.go — added ErrAutoGroupCannotBeDeleted and ErrAutoGroupMembersCannotBeModified                    
 - db/org_members_groups.go — dynamic member population on read; auto group prepended on page 1;               
 delete/member-edit blocked; ListOrganizationMemberGroup and census field validation handle auto groups        
 correctly; new EnsureAutoMemberGroup, DeleteAutoMemberGroupIfEmpty, deleteAutoMemberGroup helpers             
 - db/org_members.go — hooked auto group lifecycle into all member mutation paths (SetOrgMember,               
 AddBulkOrgMembers, UpsertOrgMemberAndCensusParticipants, DelOrgMember, DeleteOrgMembers,                      
 DeleteAllOrgMembers); fixed lock release ordering to avoid deadlocks                                          
 - errors/errors_definition.go — two new 403 Forbidden error codes                                             
 - api/apicommon/types.go — IsAutoGroup bool added to OrganizationMemberGroupInfo                              
 - api/organization_groups.go — isAutoGroup passed through in all responses; handlers return correct errors    
 for delete/update attempts on auto groups                                                                     
 - migrations/0011_create_all_members_auto_group.go — back-fills the auto group for all existing organizations 
 that already have members (idempotent)                                                                        
                                                                                                               
 ### Tests                                                                                                     
                                                                                                               
 10 new integration tests covering all acceptance criteria, plus updates to existing group tests to account    
 for the auto group now appearing in listings.

closes #434

### DIFF
--- a/db/org_members.go
+++ b/db/org_members.go
@@ -341,10 +341,10 @@ func (ms *MongoStorage) addOrgMemberBatches(
 	// Ensure the auto group exists now that members have been added.
 	if job.Added > 0 {
 		ms.keysLock.Lock()
+		defer ms.keysLock.Unlock()
 		if err := ms.EnsureAutoMemberGroup(org.Address); err != nil {
 			log.Warnw("could not ensure auto member group after bulk add", "error", err)
 		}
-		ms.keysLock.Unlock()
 	}
 }
 
--- a/db/org_members.go
+++ b/db/org_members.go
@@ -43,12 +43,17 @@ func (ms *MongoStorage) SetOrgMember(salt string, orgMember *OrgMember) (string,
 		return "", err
 	}
 	ms.keysLock.Lock()
-	defer ms.keysLock.Unlock()
 	filter := bson.M{"_id": member.ID}
 	opts := options.Update().SetUpsert(true)
 	_, err = ms.orgMembers.UpdateOne(ctx, filter, updateDoc, opts)
+	ms.keysLock.Unlock()
 	if err != nil {
 		return "", err
+	}
+
+	// Ensure the auto group exists now that at least one member is present.
+	if err := ms.EnsureAutoMemberGroup(orgMember.OrgAddress); err != nil {
+		log.Warnw("could not ensure auto member group after SetOrgMember", "error", err)
 	}
 
 	return member.ID.Hex(), nil
@@ -61,8 +66,18 @@ func (ms *MongoStorage) DelOrgMember(id string) error {
 		return ErrInvalidData
 	}
 
+	// Fetch the member first so we know which org it belongs to (needed for auto group cleanup).
+	ctxFetch, cancelFetch := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancelFetch()
+	var existing OrgMember
+	if err := ms.orgMembers.FindOne(ctxFetch, bson.M{"_id": objID}).Decode(&existing); err != nil {
+		if err == mongo.ErrNoDocuments {
+			return ErrNotFound
+		}
+		return fmt.Errorf("could not fetch member before deletion: %w", err)
+	}
+
 	ms.keysLock.Lock()
-	defer ms.keysLock.Unlock()
 	// create a context with a timeout
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
 	defer cancel()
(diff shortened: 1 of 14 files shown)


db/org_members.go
```diff
@@ -43,14 +43,19 @@ func (ms *MongoStorage) SetOrgMember(salt string, orgMember *OrgMember) (string,
 		return "", err
 	}
 	ms.keysLock.Lock()
-	defer ms.keysLock.Unlock()
```
emmdim, line 46: **I would not follow this pattern. I would leave the lock in the higher level function preferably as in the rest of the code. 
I would leave all the `defer ms.keysLock.Unlock()` in the code as they are and  I would remove the locks from inside `EnsureAutoMemberGroup`**

db/org_members.go
```diff
@@ -313,6 +337,13 @@ func (ms *MongoStorage) addOrgMemberBatches(
 			Errors:   append(job.Errors, errs...),
 		}
 	}
+
```
emmdim, line 340: If the lock comment above is implemnted the, locks need to be added here

emmdim (changes requested):
Please address te following comments. more coming

db/org_members_groups.go
```diff
@@ -41,10 +41,20 @@ func (ms *MongoStorage) OrganizationMemberGroup(
 		}
 		return nil, err
 	}
+	// For the auto group, populate MemberIDs dynamically from the full orgMembers collection.
+	if group.IsAutoGroup {
```
emmdim: `OrganizationMemberGroup` is a lighter call that does not retrieve the orgMembers (as done by `ListOrgMembersGroup`), so this code is not relevant here

db/org_members_groups.go
```diff
@@ -53,9 +63,45 @@ func (ms *MongoStorage) OrganizationMemberGroups(
 		return 0, nil, ErrInvalidData
 	}
 
-	filter := bson.M{"orgAddress": orgAddress}
+	// Fetch the auto group separately (may not exist yet).
```
emmdim: The logic here seems ok, but seems like an overkill, since you can actually maintain the older code and use a sort of the results based on first the  `isAutoGroup` (and second the groupID if needed) in order to achieve the same outcome.

What do you think?

emmdim (changes requested):

migrations/0011_create_all_members_auto_group.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	return nil
+}
+
+// downCreateAllMembersAutoGroup removes all auto groups created by this migration.
+// It is safe to run multiple times.
+func downCreateAllMembersAutoGroup(_ context.Context, _ *mongo.Database) error {
+	// We intentionally do not delete auto groups on rollback to avoid data loss:
+	// the groups are harmless and may already be referenced by census records.
```
emmdim: maybe the down part here would make sense to delete all autoGroups?

emmdim (changes requested):

migrations/0011_create_all_members_auto_group.go
```diff
@@ -0,0 +1,73 @@
...
+// organization that already has at least one member but does not yet have an
+// auto group.  New organizations receive their auto group automatically when
+// the first member is added via the application layer, so this migration only
+// needs to back-fill existing data.
+func upCreateAllMembersAutoGroup(ctx context.Context, database *mongo.Database) error {
+	orgMembers := database.Collection("orgMembers")
+	orgMemberGroups := database.Collection("orgMemberGroups")
+
+	// Find all distinct orgAddress values that have at least one member.
+	addresses, err := orgMembers.Distinct(ctx, "orgAddress", bson.D{})
```
emmdim: Cosnidering that we are talking about 100s of thousands of users maybe it would be more efficient to simply get all the organizations and extract from them the addresses?

elboletaire:
This is ready for review and can be safely squashed @emmdim. I kept the commits split for agent debugging purposes.

emmdim (approved):
LGTM

PR #466, opened 2026-05-24
fix(api): percent-encode cursor in pagination next/previous links

### LINKED ISSUE #458: Pagination cursor not URL-encoded in next-page link
**Repro:** POST 30 organizations, then `GET /organizations?limit=10`. Inspect the `next` field in the response.

**Expected:** the `cursor` query value is URL-encoded.
**Actual:** raw bytes appear in the link, breaking when cursor contains `+` or `=`.

Fix: percent-encode the cursor value when building the next link. Add a test that asserts encoding for a cursor containing `+`.

argos-code:
## Summary

Pagination responses for the `/organizations/{address}/jobs` endpoint now include `next` and `previous` URL fields in the `pagination` object. These URLs are fully percent-encoded (via `url.Values.Encode()`), so reserved characters like `+` and `=` that may appear in cursor values are never left raw.

## Linked issue

Closes #458

## Changes

- **`api/apicommon/types.go`** — Added `Next string` and `Previous string` fields to `Pagination`. Both are `omitempty`; they are only populated when a caller supplies a base URL.
- **`api/helpers.go`** — `calculatePagination` now accepts an optional `baseURL ...string` (variadic, fully backward-compatible). When a non-empty base URL is provided, `buildPageURL` constructs percent-encoded next/previous links by parsing the URL, setting `page` and `limit` via `url.Values`, and round-tripping through `url.URL.String()`.
- **`api/helpers_test.go`** *(new)* — Unit tests covering: middle page (both links populated), first page (no `previous`), last page (no `next`), and empty `baseURL` (both links empty).
- **`api/organizations.go`** — `organizationJobsHandler` passes `a.serverURL + r.URL.RequestURI()` as the base URL, preserving any existing query parameters (e.g. `?type=org_members`) while `buildPageURL` overwrites only `page` and `limit`.
- **`api/organizations_test.go`** — Integration test `TestOrganizationJobsPaginationNextLink`: seeds 15 jobs, fetches page 1 with `limit=10`, asserts `next` is non-empty, contains no raw `+`, contains `page=2` and `limit=10`; then follows the link and asserts the second page returns the remaining 5 jobs with `previous` populated and `next` empty.

## Tests run

- `go vet ./...` — clean
- `golangci-lint run ./...` — 0 issues
- `go build ./...` — clean
- Unit tests in `api/helpers_test.go` — pass (`GOTMPDIR=/workspace/gotmp go test ./api/ -run TestCalculatePagination`)
- Integration tests (`./api/...`) require testcontainers and are validated in CI

## Risks and review focus

- **Backward compatibility**: all existing callers of `calculatePagination` pass 3 arguments; the variadic fourth parameter is never required, so no call sites break.
- **Base URL construction**: using `r.URL.RequestURI()` (path + raw query) means existing filter params survive across pages. Reviewers should confirm this is the desired behaviour for endpoints that combine filtering and pagination.
- **`omitempty` on `Next`/`Previous`**: JSON consumers that previously relied on the absence of these fields will now see them. This is an additive, non-breaking change.

## Notes for maintainers

Other handlers that call `calculatePagination` without a base URL continue to work unchanged and do not emit `next`/`previous` fields. To opt in, pass `a.serverURL + r.URL.RequestURI()` as the fourth argument (same pattern used in `organizationJobsHandler`).

### DIFF
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -31,6 +31,10 @@ type Pagination struct {
 	CurrentPage  int64  `json:"currentPage"`
 	NextPage     *int64 `json:"nextPage"`
 	LastPage     int64  `json:"lastPage"`
+	// Next is the full URL (with percent-encoded cursor) to fetch the next page, empty when on the last page.
+	Next string `json:"next,omitempty"`
+	// Previous is the full URL (with percent-encoded cursor) to fetch the previous page, empty when on the first page.
+	Previous string `json:"previous,omitempty"`
 }
 
 // PaginationParams allows the client to request a specific page, and how many items per page
--- a/api/helpers.go
+++ b/api/helpers.go
@@ -210,8 +210,10 @@ func (a *API) generateVerificationLink(user *db.User, code string) (string, erro
 
 // calculatePagination calculates PreviousPage, NextPage and LastPage.
 //
-// If page is negative or higher than LastPage, returns ErrPageNotFound
-func calculatePagination(page, limit, totalItems int64) (*apicommon.Pagination, error) {
+// If page is negative or higher than LastPage, returns ErrPageNotFound.
+// The optional baseURL parameter, when non-empty, causes Next and Previous to be
+// populated with fully-formed, percent-encoded URLs (e.g. "https://host/path?limit=10&page=3").
+func calculatePagination(page, limit, totalItems int64, baseURL ...string) (*apicommon.Pagination, error) {
 	lastp := int64(math.Ceil(float64(totalItems) / float64(limit)))
 	if totalItems == 0 {
 		lastp = 1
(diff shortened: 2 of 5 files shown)


argos-code:
❌ **Not applicable**

This is an informational sticky comment automatically posted by the `swag.yml` CI workflow whenever a PR touches `api/**/*` files. The workflow regenerates the swagger docs, diffs them against the committed copy, and posts any differences here — but it deliberately does NOT commit the regenerated file back to the PR branch. On merge to main the CI auto-commits the updated `swagger.yaml`. The diff shown is correct and accurate: it reflects the new `Next` and `Previous` string fields added to `apicommon.Pagination` in `api/apicommon/types.go`. No action is required on the PR itself.

p4u (changes requested):
Could you also add a Previous link guarding the same encoding case? The test currently asserts the Next cursor is percent-encoded but doesn't verify Previous; ideally the helper that builds both links shares the same QueryEscape path, with a test that asserts both are encoded when given a cursor containing '+'. Single helper rather than two parallel code paths would be much cleaner here.

argos-code:
❌ **Not applicable**

This is an automated Coveralls report, not a review request. The two uncovered lines are the `url.Parse` error path inside `buildPageURL` (triggered only by a structurally invalid URL string). I've added a unit test case in `helpers_test.go` that passes a control-character URL (`://\x00bad`) to exercise that branch, which will improve coverage. No code change is needed for the production path.

argos-code:
✅ **Accepted**

Added Previous-link encoding assertions to the integration test and a malformed-URL unit test case covering the buildPageURL error branch.

emmdim:
Currently irrelevant

PR #485, opened 2026-05-24
test(db): add regression test for verification code expiration boundary

### LINKED ISSUE #481: argos-test: add regression test for verification-code expiration race
There appears to be a race between verification-code expiration and the SetVerificationCode call: if a code is generated exactly at the expiration boundary, the user can use the previous code. Add a unit test in `db/verifications_test.go` that exercises this boundary case and asserts the new code wins. If the test does not actually fail on current main, document why.

argos-code:
## Summary

Adds `TestSetVerificationCodeExpirationBoundary` to `db/verifications_test.go` to exercise the boundary case between verification-code expiration and `SetVerificationCode`. The test confirms the new code always wins after a `SetVerificationCode` call, even when the call lands at or after the previous code's expiration time.

The test does not fail on current `main`. An inline comment documents why: `UserVerificationCode` does not filter by expiration (no MongoDB TTL index on the collection, no application-level expiry check at query time), and `SetVerificationCode` uses an atomic `ReplaceOne` (upsert) that completely overwrites the existing document — so once the call returns, any subsequent read is guaranteed to see the new code.

## Linked issue

Closes #481

## Changes

- `db/verifications_test.go`: new subtest `TestSetVerificationCodeExpirationBoundary` inside `TestVerifications`
  - Creates a user, sets a verification code expiring in 1 ms
  - Waits 2 ms past that boundary
  - Calls `SetVerificationCode` with a new code and a 1-hour TTL
  - Asserts `UserVerificationCode` returns the new sealed code with a future expiration
  - Comment in test explains why no race is observable on current `main`

No production code was modified.

## Tests run

```
$ golangci-lint run --new-from-rev=main ./db/...
0 issues.

$ ./scripts/check-qt-patterns.sh
✅ No qt anti-patterns found!
```

The full `db/...` suite (including the new test) requires Docker/testcontainers and runs in CI.

## Risks and review focus

- The test uses `time.Sleep(2 * time.Millisecond)` for boundary timing. On heavily loaded CI runners this is unlikely to be flaky, but worth keeping in mind if the test shows intermittent failures in future.
- No production paths touched; risk is confined to test infrastructure.

## Notes for maintainers

As documented in the test comment, expiration on `UserVerification` documents is currently unenforced at query time and there is no MongoDB TTL index on the `verifications` collection (unlike `organizationInvites`, which does have one). If enforcement of expired codes is ever desired, a TTL index or an application-level filter in `UserVerificationCode` would be needed — that is a separate issue.

### DIFF
--- a/db/verifications_test.go
+++ b/db/verifications_test.go
@@ -95,4 +95,42 @@ func TestVerifications(t *testing.T) {
 		c.Assert(err, qt.IsNil)
 		c.Assert(code.Attempts, qt.Equals, 2)
 	})
+
+	t.Run("TestSetVerificationCodeExpirationBoundary", func(_ *testing.T) {
+		c.Assert(testDB.DeleteAllDocuments(), qt.IsNil)
+		userID, err := testDB.SetUser(&User{
+			Email:     testUserEmail + "4",
+			Password:  testUserPass,
+			FirstName: testUserFirstName,
+			LastName:  testUserLastName,
+		})
+		c.Assert(err, qt.IsNil)
+
+		sealedCode1, err := internal.SealToken("oldCode", testUserEmail, "mock-app-secret")
+		c.Assert(err, qt.IsNil)
+		// Set a code that expires almost immediately, simulating the expiration boundary.
+		oldExp := time.Now().Add(time.Millisecond)
+		c.Assert(testDB.SetVerificationCode(&User{ID: userID}, sealedCode1, CodeTypeVerifyAccount, oldExp), qt.IsNil)
+
+		// Wait past the old code's expiration boundary.
+		time.Sleep(2 * time.Millisecond)
+
+		// Set a new code at (or just after) the old code's expiration boundary.
+		sealedCode2, err := internal.SealToken("newCode", testUserEmail, "mock-app-secret")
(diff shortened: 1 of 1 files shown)


argos-code:
❌ **Not applicable**

This is an automated Coveralls report, not a reviewer request requiring action. Coverage remaining the same at 59.566% is the expected and correct outcome: this PR adds only test code to `db/verifications_test.go`. Test files are not counted as coverable lines by Coveralls — they are the test harness, not production code under test. The report explicitly confirms 'No coverage regressions found' and 'No uncovered changes found', which is a clean result. No changes are needed.

p4u (commented):
The boundary test passes today, which is good — but please add a t.Log noting whether the assertion would have failed on a prior commit if you can identify one, or otherwise mark the test as 'documenting current behavior' rather than 'regression for fixed bug'. Either is fine, just be honest about which.

argos-code:
✅ **Accepted**

Added t.Log and an explanatory comment block at the top of TestSetVerificationCodeExpirationBoundary explicitly marking it as 'documenting current behavior (not a regression for a fixed bug)'; removed the now-redundant mid-test comment that conveyed the same information.

PR #504, opened 2026-06-02
feat(db/census):  update census size with new participants

lucasmenendez:
update census size in the database when new census participants are added


emmdim (approved):
LGTM

PR #505, opened 2026-06-02
feat(csp): add token-only census membership check endpoint

elboletaire:
Add POST /process/bundle/{bundleId}/check so the UI can verify, using only the stored CSP auth token, whether a voter belongs to a bundle's census — mirroring client.isInCensus() for non-CSP censuses. Previously the CSP vote button was gated on mere token presence, so it was wrongly shown when the stored token belonged to a different bundle than the election being viewed.

The endpoint returns {belongs, weight, hasVoted}. belongs is true only for a verified token issued for this same bundle whose user is a participant of the bundle's census (and non-zero weight on weighted censuses). Ineligibility is reported as belongs:false with HTTP 200, not as an error, so the UI can consume it the same way it consumes isInCensus.

Composes existing DB methods (CSPAuth, OrgMember, Census, CensusParticipant, CSPProcessByUserAndProcess); no new query logic.

Required by https://github.com/vocdoni/vocdoni-app/issues/1675


emmdim (approved):
LGTM

PR #507, opened 2026-06-04
chore(stage): upgrade from main

elboletaire:


elboletaire:
Kinda nonsense that github action tbh, specially when usually we're not using the PR title as commit message.

PR #511, opened 2026-06-08
fix(members): normalize emails for case-insensitive CSP login

### LINKED ISSUE #508: Normalize emails in memberbase
Follow https://github.com/vocdoni/vocdoni-app/issues/1649

elboletaire:
Member emails were stored as-typed (mixed case). The CSP 2FA login finds a participant by recomputing a login hash that includes the email, then does an exact, case-sensitive hash match, so a member stored as `User@X.com` could not be matched by `user@x.com`. The `strings.EqualFold` check in verifyEmail runs only after the hash lookup already succeeded, so it never helped with case.

Normalize at the write chokepoint and at login, and backfill existing data:

- internal: add NormalizeEmail (trim + lowercase) and extract the login-hash primitive into HashSortedFields so the migration and db share one implementation. The sha256.New().Sum(fmt.Append(...)) construction is the legacy on-disk format and must not change.
- db: refactor HashAuthTwoFaFields to use the shared primitive; lowercase the email in prepareOrgMember (the single write path all member inserts/upserts and bulk/CSV imports funnel through), so stored emails and the hashes derived from them are canonical.
- csp: lowercase the login email input in authFirstStep so the lookup hash matches the now-normalized stored hash.
- migrations: add 0015 to lowercase existing member emails and recompute the affected participant login hashes. Members whose lowercased email would collide on a census login hash (unique index from 0008) are skipped and logged for manual review instead of aborting the run. Irreversible by design.

Tests: migration normalization + hash recompute + end-to-end login lookup, collision skip, and write-path normalization.

closes #508

### DIFF
--- a/db/org_members.go
+++ b/db/org_members.go
@@ -171,8 +171,11 @@ func prepareOrgMember(org *Organization, m *OrgMember, salt string, currentTime 
 	}
 	member.OrgAddress = org.Address
 
-	// check if mail is valid
+	// normalize and validate the email. Normalizing to lowercase here is the
+	// single write chokepoint that keeps every member email (and the login
+	// hashes derived from it) case-insensitive across all write paths.
 	if member.Email != "" {
+		member.Email = internal.NormalizeEmail(member.Email)
 		if _, err := mail.ParseAddress(member.Email); err != nil {
 			errors = append(errors, fmt.Errorf("could not parse email: %s %v", member.Email, err))
 			// If email is invalid, set it to empty and store the error
--- a/db/org_members.go
+++ b/db/org_members.go
@@ -171,11 +171,9 @@ func prepareOrgMember(org *Organization, m *OrgMember, salt string, currentTime 
 	}
 	member.OrgAddress = org.Address
 
-	// normalize and validate the email. Normalizing to lowercase here is the
-	// single write chokepoint that keeps every member email (and the login
-	// hashes derived from it) case-insensitive across all write paths.
+	// normalize and validate the email
+	member.Email = internal.NormalizeEmail(member.Email)
 	if member.Email != "" {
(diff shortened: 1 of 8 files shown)


db/org_members.go
```diff
@@ -171,8 +171,11 @@ func prepareOrgMember(org *Organization, m *OrgMember, salt string, currentTime
 	}
 	member.OrgAddress = org.Address
 
-	// check if mail is valid
+	// normalize and validate the email. Normalizing to lowercase here is the
+	// single write chokepoint that keeps every member email (and the login
+	// hashes derived from it) case-insensitive across all write paths.
 	if member.Email != "" {
+		member.Email = internal.NormalizeEmail(member.Email)
```
lucasmenendez: Maybe this line should be moved before the `if` (the previous one), since the normalize function trims spaces.

migrations/0015_normalize_member_emails.go
```diff
@@ -0,0 +1,272 @@
...
+
+	log.Infow("normalized member emails", "normalized", normalized, "skipped", skipped)
+	return nil
+}
+
+// downNormalizeMemberEmails is intentionally a no-op: the normalization is lossy
+// (the original mixed-case emails and their derived login hashes cannot be
+// reconstructed), so this migration is irreversible by design.
+func downNormalizeMemberEmails(_ context.Context, _ *mongo.Database) error {
+	return nil
```
lucasmenendez, line 272: I suppose that this migration is one-way right?

lucasmenendez (dismissed):

migrations/0015_normalize_member_emails.go
elboletaire: Yup, there's no way back

altergui (approved):

PR #520, opened 2026-06-11
refactor(notifications): unify API email sends with CSP queue infrastructure

### LINKED ISSUE #522: "Recover your password" email has no cooldown or retry logic
recoverUserPasswordHandler uses a plain a.sendMail() call, allowing spam

altergui:
## Summary

- **Extracts queue infrastructure** (`Breaker`, `Queue`, `QueueItem`) from `csp/notifications/` into the top-level `notifications/` package so both the CSP and the API can share it
- **Fixes two rogue direct sends** in the API: `recoverUserPasswordHandler` (via `sendMail`) and `organizationCreateTicket` (was calling `a.mail.SendNotification` directly) — both now go through the queue with async delivery, retry-on-transient, and circuit-breaking
- **Adds a 60-second cooldown** to `POST /users/password/recovery` via a new `UserVerification.CreatedAt` field, matching the rate-limiting intent of the CSP OTP system

## What changed

| Area | Change |
|------|--------|
| `notifications/breaker.go` | Moved + exported (`Breaker`, `NewBreaker`) from `csp/notifications/` |
| `notifications/queue.go` | New generic `Queue` / `QueueItem` with workers, circuit-breaker, TTL, retry |
| `csp/notifications/queue.go` | Thin wrapper: maps `NotificationChallenge ↔ QueueItem` via `Meta`, feeds `NotificationsSent` from inner `Done` channel — external interface unchanged |
| `api/api.go` | Added `notifyQueue *notifications.Queue`, init in `New()`, start in `Start()` |
| `api/notifications.go` | `sendMail` now enqueues (fire-and-forget); new `queueNotification` helper |
| `api/organizations.go` | Support-ticket send → `queueNotification` |
| `api/users.go` | 60-second cooldown check before issuing new password reset code |
| `db/types.go` + `db/verifications.go` | `UserVerification.CreatedAt` field |

## Test plan

- [x] `make test` — full integration test suite (spins Mongo + Voconed via testcontainers)
- [ ] `POST /users/password/recovery` for a valid email → 200, email arrives
- [ ] Repeat within 60 s → 200 silently, no second email sent
- [ ] `POST /organizations/{addr}/tickets` → 200, support email arrives
- [ ] With SMTP blocked: both flows return 200 immediately; emails deliver once SMTP recovers (retry)
- [ ] CSP OTP flow unaffected (same external interface, integration tests cover it)

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -47,6 +47,7 @@
 package api
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"time"
@@ -93,6 +94,12 @@ type Config struct {
 	CSP           *csp.CSP
 	// OAuth service URL
 	OAuthServiceURL string
+	// OTPExpiry overrides the validity window for all one-time codes (account
+	// verification, password reset). Zero uses notifications.DefaultOTPExpiry.
+	OTPExpiry time.Duration
+	// OTPCooldown overrides the anti-spam rate limit between notification
+	// requests for the same account. Zero uses notifications.DefaultOTPCooldown.
+	OTPCooldown time.Duration
 }
 
 // API type represents the API HTTP server with JWT authentication capabilities.
@@ -106,6 +113,8 @@ type API struct {
 	account         *account.Account
 	mail            notifications.NotificationService
 	sms             notifications.NotificationService
+	ctx             context.Context
+	notifyQueue     *notifications.Queue
 	secret          string
 	webAppURL       string
 	serverURL       string
@@ -117,10 +126,12 @@ type API struct {
 	stripeHandlers  *StripeHandlers
 	txQueue         chan txTask
 	orgTxLocks      *orgTxMutex
+	otpExpiry       time.Duration
+	otpCooldown     time.Duration
 }
 
 // New creates a new API HTTP server. It does not start the server. Use Start() for that.
-func New(conf *Config) *API {
+func New(ctx context.Context, conf *Config) *API {
 	if conf == nil {
 		return nil
 	}
@@ -129,7 +140,25 @@ func New(conf *Config) *API {
 		conf.ObjectStorage.ServerURL = conf.ServerURL
 	}
 
+	otpExpiry := conf.OTPExpiry
+	if otpExpiry <= 0 {
+		otpExpiry = notifications.DefaultOTPExpiry
+	}
+	otpCooldown := conf.OTPCooldown
+	if otpCooldown <= 0 {
+		otpCooldown = notifications.DefaultOTPCooldown
+	}
+
+	var notifyQueue *notifications.Queue
+	if conf.MailService != nil || conf.SMSService != nil {
+		notifyQueue = notifications.NewQueue(ctx, notifications.QueueConfig{
+			MailService: conf.MailService,
+			SMSService:  conf.SMSService,
+		})
+	}
+
 	a := &API{
+		ctx:             ctx,
 		db:              conf.DB,
 		auth:            jwtauth.New("HS256", []byte(conf.Secret), nil),
 		host:            conf.Host,
@@ -138,6 +167,7 @@ func New(conf *Config) *API {
 		account:         conf.Account,
 		mail:            conf.MailService,
 		sms:             conf.SMSService,
+		notifyQueue:     notifyQueue,
 		secret:          conf.Secret,
 		webAppURL:       conf.WebAppURL,
 		serverURL:       conf.ServerURL,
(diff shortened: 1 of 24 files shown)


emmdim (approved):

emmdim:
FYI @altergui  apparently this PR introduces a race that was triggered after rebasing the main branch:

## What's happening

The PR changed `sendMail` from **synchronous** to **fire-and-forget through the shared `notifyQueue`**:

- **Before the PR:** `sendMail` blocked on `SendNotification`, so the verification email was in the fake SMTP mailbox *before the register request returned*. `waitForEmail` always found it immediately. Guaranteed happens-before.
- **After the PR:** `sendMail` just `Push`es to the queue and returns. Delivery happens later on a worker goroutine. `waitForEmail` (10 polls × ~3s ≈ 30s budget) now **races the queue worker**. The `EOF` is just `FindEmail`'s "not there yet."

So every `testCreateUser` now depends on the queue draining within ~30s instead of being synchronous.

## Why pre-rebase CI passed, post-rebase fails

Async-alone was fast enough to win the race on the PR branch. The rebase pulled in `main`'s heavier concurrent work (async tx job queue + its worker goroutines, on-chain provisioning, integrator atomics) all competing for CI's 2 slow cores. Two ways that tips it over (`notifications/queue.go` + `breaker.go`):

1. **Plain latency:** worker delivery slips past the 30s poll window under CPU starvation.
2. **Breaker stall (worse):** each `SendNotification` has a 10s timeout. Under load, sends to the fake SMTP time out → counted as transient failures → after 10 the mail breaker opens for 30s (`DefaultBreakerCooldown`). During that window every email is `softReenqueue`d and skipped (`br.Allow()` false) — so user-2's verification email is gone for the *entire* `waitForEmail` budget.

The failing call being the **second** `testCreateUser` (right after `testCreateOrganization`'s extra work) fits both.

## Fix options (minimal → proper)

- **Band-aid:** widen `waitForEmail` budget. Hides the race, doesn't fix it.
- **Test-correct:** in `TestMain`, configure the queue with `Workers: 1`-ish but more importantly disable the breaker for tests (`BreakerMaxFailures` very high) and bump the send timeout — removes the 30s breaker stall. Still racy on pure latency.
- **Proper:** don't route OTP/verification emails (which a caller is actively waiting on) through the same best-effort, breaker-gated queue as fan-out mail — or have the queue expose a per-item `Done` that tests (and the register handler in test mode) can await.

altergui:
Thanks @emmdim — spot-on analysis. Fixed by making delivery *awaitable* rather than widening the poll budget, and rebased onto fresh `main`.

**Rebase:** rebased onto current `main` (which now includes the merged #517 CSP OTP hardening). Resolved the `csp/auth.go` + `csp/csp.go` conflicts so both land together: #517's per-token random secret / attempt-limiting / constant-time / empty-secret handling **and** this PR's unified queue + `NotificationTTL` lifecycle naming. Branch is now even with `main`, no conflicts.

**Race fix** (new commit `fix(notifications): expose per-item delivery await; remove test race`):
- `notifications.Queue` gains an optional per-item `done` channel and `PushWait(item)` returning a `<-chan` signalled exactly once when the item reaches its terminal outcome (delivered or given up).
- `api.Config.NotificationsSyncDelivery`: when set, `sendMail` blocks on `PushWait` until delivery completes (bounded by request ctx + 30s cap). **Off in production** — fire-and-forget is unchanged.
- The API test harness enables it, so a handler that sends mail then returns guarantees the mail is in the fake inbox before the HTTP response is observed. This restores the pre-PR happens-before deterministically — no reliance on the worker winning a 30s poll race, and immune to the breaker-stall path you described.
- Added `Queue.PushWait` unit tests (success + give-up).

Full `./api/...` suite passes locally (277s), plus `./notifications/`, `./csp/...`, `./db/`. Watching CI now.

altergui:
Follow-up: CI surfaced `TestRelayVote` failing on email reception — that path waits on the **CSP OTP challenge** email, which flows through the CSP's own queue (separate from the API's), so my first commit (API-only sync delivery) didn't cover it.

Extended the same per-item-await mechanism to the CSP:
- `csp/notifications.Queue` wrapper gains `PushWait` (delegates to the inner generic queue).
- `csp.Config.SyncDelivery`; when set, `BundleAuthToken`/`ResendChallenge` block until the challenge is delivered (bounded by ctx + 30s). Off in production.
- API test harness enables it on `testCSP`.

Full `./api/...` suite green locally (264s), plus `./csp/...`, `./notifications/`, `./db/`. Folded into the race-fix commit; force-pushed. Watching CI.

PR #521, opened 2026-06-11
fix(users): send verification email on password recovery for unverified accounts

altergui:
## Summary

- Unverified users who called `POST /users/password/recovery` received a silent `200 OK` with no email, leaving them unable to log in **or** recover their account (deadlock: login requires `Verified=true`, recovery silently no-ops for unverified users)
- Fix: if the account exists but is not verified, re-send the account verification email so the user can complete registration
- Email enumeration protection is preserved — non-existent addresses still return `200 OK` with no email sent

## Test plan

- [x] Register a new account, do **not** verify it
- [x] Call `POST /users/password/recovery` with that email → should receive the account verification email (not a password reset email)
- [x] Verify the email using the code/link received → account becomes verified
- [x] Confirm normal password recovery now works for the verified account
- [x] Call `POST /users/password/recovery` with an unknown email → `200 OK`, no email (enumeration protection intact)

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -578,27 +578,46 @@ func (a *API) recoverUserPasswordHandler(w http.ResponseWriter, r *http.Request)
 		errors.ErrGenericInternalServerError.Write(w)
 		return
 	}
-	// check the user is verified
-	if user.Verified {
-		// generate a new verification code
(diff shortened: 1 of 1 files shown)


emmdim (approved):
LGTM

PR #523, opened 2026-06-15
feat(api): new org bundles endpoint

### LINKED ISSUE #518: Add new endpoint to list org process related to their bundles
Add a paginated `GET /organizations/{address}/processes` endpoint that lists the bundles for an organisation with their main process (which identifies a set of processes that form a multi-process setup).
The endpoint should return a paginated list of `{bundleId, mainProcessId}` entry per bundle.

It will be required to solve the epic https://github.com/vocdoni/vocdoni-app/issues/1685.

lucasmenendez:
### Goal                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                     
Provide a paginated API endpoint that allows organization managers/admins to retrieve the list of process bundles belonging to their organization, exposing the bundle ID, its primary (first) process, and all associated process IDs.                                                              
                                                                                                                                                                                                                                                                                                     
### Summary                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                     
This PR introduces the full stack (types, database method, API handler, route, and tests) for a new read‑only endpoint `GET /organizations/{address}/processes` that returns a paginated list of process bundles for a given organization. It also refactors the existing `HasRoleFor` method on the `User` type by extracting a reusable `RoleFor` helper that returns the user's role for an organization, which the new handler uses to authorize the request.                                                                                                                                                 
                                                                                                                                                                                                                                                                                                     
### Main Changes                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                     
1. New API types (`api/apicommon/types.go`) 
    - Added `OrganizationBundle` struct: holds `bundleId`, `primaryProcessId` (the first process in the bundle), and `processes` (all process IDs in the bundle).
    - Added `ListOrganizationBundles` struct: wraps a `Pagination` object and a slice of `OrganizationBundle`.
2. New DB pagination method (`db/process_bundles.go`).
    - Added `ListOrganizationBundles(orgAddress, page, limit)`: retrieves process bundles filtered by organization address using the existing `paginatedDocuments` generic.
  - Validates that the organization address is non‑zero; returns `ErrInvalidData` otherwise.                                                                                                                                                                                                                                                                                                     
3. New HTTP handler and route (`api/organizations.go`, `api/routes.go`, `api/api.go`).
    - Registered `GET /organizations/{address}/processes` → `organizationBundlesHandler`.                                                                                                                                                                                                                    
    - The handler:
      - Authenticates the user via bearer token.
      - Resolves the organization from the request context.
      - Authorizes the request using the new `RoleFor` helper (requires the user to have a role in the organization).
      - Parses `page`/`limit` query parameters.
      - Calls `db.ListOrganizationBundles`, calculates pagination metadata, maps DB bundles to the API `OrganizationBundle` type, and writes the JSON response.
4. Refactored user role lookup (db/types.go)
    - Extracted `RoleFor(address) (UserRole, bool)` from the existing `HasRoleFor` method.
    - `HasRoleFor` now delegates to `RoleFor` and checks the specific role match.
    - `RoleFor` returns the user's role for a given organization, enabling the new handler to do a simple membership check rather than a specific‑role check.
5. Tests (`api/organizations_test.go`, `db/process_bundles_test.go`):
    - API test: `TestOrganizationBundlesHandler` creates two test bundles, verifies paginated response shape, checks authorization failures (no token, non‑member user), malformed pagination params, and non‑existent org address.
    - DB test: `TestListOrganizationBundles` creates bundles across multiple organizations, verifies correct filtering, pagination (page size/offset), empty results for orgs without bundles, and invalid address validation.
    - Added `testCreateOrganizationBundle` test helper for setting up prerequisites (census, process, bundle).

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -248,6 +248,7 @@ func (a *API) initRouter() http.Handler {
 		handle(r, http.MethodDelete, organizationGroupEndpoint, a.deleteOrganizationMemberGroupHandler)
 		handle(r, http.MethodPost, organizationGroupValidateEndpoint, a.organizationMemberGroupValidateHandler)
 		handle(r, http.MethodGet, organizationJobsEndpoint, a.organizationJobsHandler)
+		handle(r, http.MethodGet, organizationBundlesEndpoint, a.organizationBundlesHandler)
 		handle(r, http.MethodPost, subscriptionsCheckout, a.stripeHandlers.CreateSubscriptionCheckout)
 		handle(r, http.MethodGet, subscriptionsCheckoutSession, a.stripeHandlers.GetCheckoutSession)
 		handle(r, http.MethodGet, subscriptionsPortal, func(w http.ResponseWriter, r *http.Request) {
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -269,6 +269,27 @@ type ListOrganizationProcesses struct {
 	Processes []db.Process `json:"processes"`
 }
 
+// OrganizationBundle represents an organization bundle. It contains the bundle ID and the main process ID.
+// swagger:model OrganizationBundle
+type OrganizationBundle struct {
+	// The ID of the bundle
+	BundleID string `json:"bundleId"`
+	// The ID of the primary process which identifies the set of processes in
+	// the bundle, no matter how many (one or more)
+	PrimaryProcessID string `json:"primaryProcessId"`
+	// The list of processes IDs in the bundle
+	Processes []string `json:"processes"`
+}
+
+// ListOrganizationBundles represents the response for listing the bundles of an organization.
+// swagger:model ListOrganizationBundles
+type ListOrganizationBundles struct {
+	// Pagination fields
+	Pagination *Pagination `json:"pagination"`
+	// List of organization bundles
+	Bundles []OrganizationBundle `json:"bundles"`
+}
+
 // ValidateMemberGroupRequest validates the request for creating or updating an organization member group.
 // Validates that either AuthFields or TwoFaFields are provided and checks for duplicates or empty fields.
 // swagger:model ValidateMemberGroupRequest
--- a/api/organizations.go
+++ b/api/organizations.go
@@ -547,8 +547,9 @@ func (a *API) organizationJobsHandler(w http.ResponseWriter, r *http.Request) {
 //
 //	@Summary		List organization bundles and their main process
 //	@Description	Returns a paginated list of process bundles for an organization.
-//	@Description	Each entry contains the bundle ID and the primary vochain process ID (Processes[0]).
-//	@Description	Requires Manager or Admin role.
+//	@Description	Each entry contains the bundle ID, the primary vochain process ID
+//	@Description	(first element of Processes when present), and all process IDs in the bundle.
+//	@Description	Requires the user to be part of the organization (any role).
 //	@Tags			process
 //	@Accept			json
 //	@Produce		json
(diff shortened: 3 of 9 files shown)


elboletaire (changes requested):
Nice work — the read endpoint is well-structured and the tests cover the happy paths and auth gates cleanly. Reviewing against the pagination slice of #1685 (Tasks 8–11 are correctly out of scope here, tracked separately), I found **one real functional bug** to fix before merge and a couple of quick polish items.

The following are **intentional by design** (not flagging them as issues, just confirming): any-role org auth (only non-members excluded), the `OrganizationBundle`/`primaryProcessId`/`Processes[]` response contract, and empty bundles being withheld from the client. ✅

## 🐛 Issues

### 1. `TotalItems` counts empty bundles, but the handler drops them — pagination metadata is wrong
**`db/process_bundles.go` (`ListOrganizationBundles`) + `api/organizations.go` (`organizationBundlesHandler`)**

`paginatedDocuments` computes `TotalItems` via `CountDocuments({orgAddress})`, which counts **all** bundles — including empty ones. The handler then does `if len(bundle.Processes) == 0 { continue }` *after* pagination. As a result:

- `TotalItems` / `LastPage` / `NextPage` are computed over a superset of what's ever returned.
- A page can come back with fewer than `limit` items.
- A client following `NextPage` can land on a page that returns `bundles: []`.

Empty bundles are a real, reachable state (`SetProcessBundle` can create a bundle before `AddProcessesToBundle` populates it), so this will actually bite.

**Fix:** filter at the DB layer so the count and the returned set agree, then the handler-side `continue` becomes a redundant safety net:

```go
filter := bson.M{
    "orgAddress":   orgAddress,
    "processes.0":  bson.M{"$exists": true}, // has at least one process
}
return paginatedDocuments[ProcessesBundle](ms.processBundles, page, limit, filter, options.Find()...)
```

### 2. Pagination isn't deterministic — no sort applied
**`db/process_bundles.go` (`ListOrganizationBundles`)**

The call passes a bare `options.Find()`; `paginatedDocuments` only sets `Skip`/`Limit`, never a sort. MongoDB doesn't guarantee a stable order without an explicit sort, so across separate page requests bundles can be skipped or duplicated (and `UpdateOne` operations like `AddProcessesToBundle` can change natural order). The sibling paginated method `Jobs` deliberately sets a sort; this one should too.

**Fix:**

```go
options.Find().SetSort(bson.D{{Key: "_id", Value: 1}})
```

Edit: if we have a date value in the bundles, we better sort by that date... ids can return weirdly sorted results when they're not incremental.

### 3. `docs/swagger.yaml` is stale
The new route/types aren't present in `docs/swagger.yaml` (0 matches for `OrganizationBundle` / `primaryProcessId` / the route). Per `CLAUDE.md`, `make swagger` should be run and the regenerated file committed after editing handlers/types.

Tip: when the github-actions bot comments about the swagger changes, it means you forgot to do it on your side and commit those changes (don't ask why is this way, I have no idea).

### 4. Test gap — the real bugs above pass current tests
`TestOrganizationBundlesHandler` / `TestListOrganizationBundles` only create bundles **with** processes, so neither the empty-bundle count inconsistency nor the unstable ordering is exercised. Once #1 and #2 are fixed, it'd be worth adding a case with a mix of empty + non-empty bundles across >1 page, asserting that `pagination.TotalItems` equals the number of bundles actually received and that paging is stable.

elboletaire (approved):
LGTM

Issue #527, opened 2026-06-20
Codex 5.5 code audit


p4u:
### 1. High - Password reset succeeds when the reset code is wrong

Location: `api/users.go:678-694`

`resetUserPasswordHandler` writes `ErrUnauthorized` when the supplied reset code does not match the stored code, but it does not return. Execution continues and the handler stores the attacker-supplied new password.

Impact: public account takeover for any verified account with a live password-reset record. An attacker can request password recovery for a victim email, then call the reset endpoint with any code and a chosen password before the reset token expires.

Recommendation:

- Return immediately after the code mismatch error.
- Invalidate/delete the reset token after successful use.
- Add failed-attempt limiting for reset codes.
- Add regression tests for wrong code, expired code, reused code, and successful code paths.

### 2. High - OAuth signature verification does not bind signatures to the expected address

Location: `account/signature.go:60-73`; callers in `api/auth.go:197`, `api/auth.go:220`, `api/auth.go:324`, `api/auth.go:345`

`VerifySignature` recovers an address from the signature, but the check is inverted and format-inconsistent:

```go
if strings.EqualFold(calcAddress.String(), util.TrimHex(address)) {
    return fmt.Errorf("signature verification failed")
}
return nil
```

`calcAddress.String()` includes the hex prefix while `util.TrimHex(address)` removes it, so matching addresses are not reliably detected. More importantly, the function returns success when the recovered address does not match the expected address.

Impact: OAuth registration and linking can accept signatures produced by the wrong key/address. This undermines both the user-signature check and the OAuth-service-signature check. For new OAuth accounts, this can allow registration of verified accounts without a valid OAuth service signature.

Recommendation:

- Normalize both addresses the same way.
- Return an error when the recovered address differs from the expected address.
- Add tests for valid signature, wrong signer/address, malformed signature, and `0x`/non-`0x` address inputs.

### 3. High - CSP 2FA challenge codes are deterministic and do not expire

Location: `csp/auth.go:232-249`, `csp/auth.go:191-224`, `csp/auth.go:265-271`, `csp/auth.go:315-322`, `db/csp.go:17-25`, `db/csp.go:103-123`

CSP challenge codes are generated as `HOTP(sha256(userID || bundleID), counter=0)`. No server secret, per-token random challenge secret, expiry, or attempt counter is stored. Verification recomputes the same deterministic code from the token's user ID and bundle ID.

[...]

PR #528, opened 2026-06-22
fix(api): reject password reset on wrong code (#527.1)

altergui:
## Summary

Fixes issue **#527 (1) — High: Password reset succeeds when the reset code is wrong**.

`resetUserPasswordHandler` (`api/users.go`) wrote `ErrUnauthorized` on a reset-code mismatch but **did not return**. Execution fell through and stored the attacker-supplied new password — public account takeover for any verified account with a live password-reset record.

## Changes

- Return immediately after the code-mismatch error.
- Invalidate (delete) the password-reset code after a successful reset so it cannot be reused.
- Add `DeleteUserVerificationCode` DB helper (`db/verifications.go`).
- Add regression test `TestResetPasswordWrongCodeRejected` covering: wrong code rejected + password untouched, successful reset, and reused-code rejected.

## Testing

- `go test -run 'TestRecoverAndResetPassword|TestResetPasswordWrongCodeRejected' ./api/` ✅
- `golangci-lint` clean on changed files.

Refs #527

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -713,16 +713,21 @@ func (a *API) resetUserPasswordHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// invalidate the reset code before updating the password so it cannot be
+	// reused. fail closed: if the code cannot be deleted we must not report
+	// success, otherwise the still-valid code could be replayed.
+	if err := a.db.DeleteUserVerificationCode(user, db.CodeTypePasswordReset); err != nil {
+		log.Warnw("could not delete password reset code", "error", err)
+		errors.ErrGenericInternalServerError.Write(w)
+		return
+	}
+
 	// hash and update the new password
 	user.Password = internal.HexHashPassword(passwordSalt, userPasswords.NewPassword)
 	if _, err := a.db.SetUser(user); err != nil {
 		log.Warnw("could not update user password", "error", err)
 		errors.ErrGenericInternalServerError.Write(w)
 		return
-	}
-	// invalidate the reset code so it cannot be reused
-	if err := a.db.DeleteUserVerificationCode(user, db.CodeTypePasswordReset); err != nil {
-		log.Warnw("could not delete password reset code", "error", err)
 	}
 	apicommon.HTTPWriteOK(w)
 }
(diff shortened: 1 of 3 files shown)


api/users.go
```diff
@@ -691,5 +692,9 @@ func (a *API) resetUserPasswordHandler(w http.ResponseWriter, r *http.Request) {
 		errors.ErrGenericInternalServerError.Write(w)
 		return
 	}
+	// invalidate the reset code so it cannot be reused
+	if err := a.db.DeleteUserVerificationCode(user, db.CodeTypePasswordReset); err != nil {
+		log.Warnw("could not delete password reset code", "error", err)
+	}
```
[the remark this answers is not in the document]
altergui: Good catch. Reworked to fail closed: the reset code is now deleted *before* the password is updated, and a deletion failure returns 500 instead of being only logged. The handler can no longer report success while a still-valid, replayable code remains.

api/users_test.go
```diff
@@ -203,6 +204,86 @@ func TestRecoverAndResetPassword(t *testing.T) {
...
 }
 
+// TestResetPasswordWrongCodeRejected ensures that a password reset is rejected
+// when the supplied code is wrong, that the password is left untouched, and that
+// a valid code cannot be reused after a successful reset.
+func TestResetPasswordWrongCodeRejected(t *testing.T) {
+	c := qt.New(t)
+
+	// Register a user with a unique email so the test is independent
+	mail := fmt.Sprintf("%d-reset@test.com", internal.RandomInt(100000000000))
```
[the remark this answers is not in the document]
altergui: Fixed. Lowered the bound to internal.RandomInt(100000), which is 32-bit safe and still gives ample uniqueness for the test email. This also matches the bound used elsewhere in the api tests.

api/apicommon/types.go
```diff
@@ -353,11 +353,6 @@ type UserInfo struct {
 
 	// Organizations the user belongs to
 	Organizations []*UserOrganization `json:"organizations"`
-
-	// Whether the user belongs to an integrator organization (integrators have a
-	// single org). Lets clients learn integrator status straight from the login
-	// flow without an extra request. Populated by GET /users/me.
-	IsIntegrator bool `json:"isIntegrator"`
```
emmdim, line 356: @altergui  why is this removed in this PR?

altergui: good catch

altergui: Restored. The field removal was a stale-base artifact — `isIntegrator` landed on `main` via a separate PR, and this branch had been cut before that. Rebased on fresh `main`, so `UserInfo.IsIntegrator` is preserved (now line 356) and the PR diff is limited to the password-reset fix.

emmdim (approved):
LGTM

PR #529, opened 2026-06-22
fix(account): bind OAuth signature verification to expected address (#527.2)

altergui:
## Summary

Fixes issue **#527 (2) — High: OAuth signature verification does not bind signatures to the expected address**.

`account.VerifySignature` had two defects:
1. **Inverted check** — it returned success when the recovered address did **not** match the expected one (`if strings.EqualFold(...) { return error }`).
2. **Format mismatch** — `calcAddress.String()` keeps the `0x` prefix while the expected address was passed through `util.TrimHex`, so even matching addresses compared unequal.

This allowed OAuth registration and account linking (`api/auth.go:197,220,324,345`) to accept signatures produced by the wrong key/address, enabling registration of verified accounts without a valid OAuth-service signature.

## Changes

- Recover the signer address and compare it against `common.HexToAddress(address)`, which normalizes the `0x` prefix and letter case reliably.
- Return an error when the recovered address differs from the expected one.
- Add `account/signature_test.go` covering valid signature, wrong signer, tampered message, malformed/empty signature, and `0x` / non-`0x` / lowercase address inputs.

## Testing

- `go test ./account/ -run TestVerifySignature -v` ✅ (all subtests pass)
- `golangci-lint` / `gofumpt` clean.

Refs #527

### DIFF
--- a/account/signature.go
+++ b/account/signature.go
@@ -3,8 +3,8 @@ package account
 import (
 	"encoding/hex"
 	"fmt"
-	"strings"
 
+	"github.com/ethereum/go-ethereum/common"
 	"go.vocdoni.io/dvote/crypto/ethereum"
 	"go.vocdoni.io/dvote/util"
 	"go.vocdoni.io/proto/build/go/models"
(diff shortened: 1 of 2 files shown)


emmdim (approved):
LGTM

PR #532, opened 2026-06-22
feat(integrator): free tier on org creation + Stripe integrator-plan support

altergui:
## What

Two related changes for self-serve integrators (combined into one PR for a single review). Builds on #531 (plan-derived integrator status, merged).

### 1. Free integrator tier on org creation
- Seed a zero-priced **"Free Integrator"** plan (migration `0012`): `MaxManagedOrgs: 1`, `MaxManagedProcesses: 5`, `MaxManagedCensusSize: 100`. `$setOnInsert` so it seeds once and stays tunable.
- **`db.FreeIntegratorPlan()`** resolves it (free plan granting managed-org capacity).
- **`POST /organizations` accepts `integrator: true`** → subscribes the new org to the free plan (active) instead of the default. The creator becomes admin and the org is an integrator immediately — **no checkout**.

### 2. Stripe integrator-plan support (paid tiers)
- **`processProductToPlan`** reads an optional `integratorLimits` Stripe **product-metadata** JSON into `Plan.IntegratorLimits` (optional — non-integrator products sync unchanged).
- **`GET /plans`** exposes `integratorLimits` via a new `SubscriptionIntegratorLimits` mirror, so clients can identify integrator plans for checkout.

Together: free tier needs no Stripe; paid upgrade = subscribe to a Stripe plan whose `integratorLimits.maxManagedOrgs > 0` via the existing embedded checkout, and (with #531) the org becomes an integrator with the plan's limits.

## Stripe product config (for the agent creating the products)
Add a **product metadata** key `integratorLimits` (same convention as `organization`/`votingTypes`/`features`):
```json
{"maxManagedOrgs": 25, "maxManagedProcesses": 250, "maxManagedCensusSize": 5000}
```
Products still need both a monthly and a yearly recurring price. The **free tier is not a Stripe product** — it's the seeded plan above.

## Design notes for review (@p4u)
1. **Free-tier quota** `1 / 5 / 100` — tune as you like (migration + in-DB).
2. **Seeded plan id** `9001` (sentinel; won't collide with Stripe's small config-index ids since `nextPlanID` is `max(_id)+1`).
3. **`FreeIntegratorPlan` selector** identifies by "free + grants managed orgs" rather than a hardcoded id.

## Verification
- `go build ./...`, `go vet`, `go test ./stripe/... ./subscriptions/...` ✅
- `db`/`api`/`migrations` test packages compile; `TestPlans/FreeIntegratorPlan` + stripe metadata-parsing tests added.

## Frontend
Pairs with `vocdoni/vocdoni-integrator-app`: org creation sends `integrator: true` (free tier, live); the embedded-checkout upgrade UI will use `GET /plans` + `POST /subscriptions/checkout` once the Stripe products + publishable key exist.

(Supersedes #533, folded in here.)

### DIFF
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -560,6 +560,9 @@ type SubscriptionPlan struct {
 
 	// Features available in this plan
 	Features SubscriptionFeatures `json:"features"`
+
+	// Integrator limits for this plan (zero when the plan is not an integrator plan)
+	IntegratorLimits SubscriptionIntegratorLimits `json:"integratorLimits"`
 }
 
 // SubscriptionPlanFromDB converts a db.Plan to a SubscriptionPlan.
@@ -579,6 +582,7 @@ func SubscriptionPlanFromDB(plan *db.Plan) SubscriptionPlan {
 		Organization:         SubscriptionPlanLimits(plan.Organization),
 		VotingTypes:          SubscriptionVotingTypes(plan.VotingTypes),
 		Features:             SubscriptionFeatures(plan.Features),
+		IntegratorLimits:     SubscriptionIntegratorLimits(plan.IntegratorLimits),
 	}
 }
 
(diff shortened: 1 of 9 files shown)


altergui:
Fixed the failing test (and a related issue it surfaced):

- **`TestPlans/FreeIntegratorPlan`** assumed an empty `plans` collection, but `DeleteAllDocuments` deliberately **preserves `plans`** and migration `0012` seeds the free integrator plan — so it's never empty. Reworked it to seed an explicit free integrator plan and assert the selector returns a zero-priced, managed-org-granting plan (robust to `FindOne` ordering). (`7e4971d`)

- While verifying, I noticed the seeded `$0` free integrator plan would otherwise show up in **public `GET /plans`** — i.e. on the main app's pricing page, and it would break `TestPlansAPI`'s `HasLen 3`. Since that plan is auto-assigned at org creation and isn't purchasable, `plansHandler` now **filters it out** of the public listing (paid integrator plans stay visible for the portal). (`26c93dc`)

@p4u — flagging the `plansHandler` filter as a small design choice: it hides only the *free* integrator plan (zero-priced + grants managed orgs) from the public list. If you'd prefer an explicit `hidden`/`internal` flag on `Plan` instead of the price-based predicate, easy to switch.

Branch now: free-tier → Stripe-limits sync → test fix → plans-listing fix. `go build`/`vet`/`go test ./stripe/... ./subscriptions/...` green locally; the DB/api integration tests run in CI.

lucasmenendez (approved):

PR #535, opened 2026-06-22
feat(apikeys): scoped API keys for organizations

altergui:
## What

Adds **API keys** so integrators can call the API programmatically (no JWT session). Per-key **scopes**, **revoke**, **optional expiry**, and **last-used** tracking.

> ⚠️ Touches the core auth middleware — flagging @emmdim @lucasmenendez. Design summary below; happy to adjust the scope taxonomy or the enforcement approach.

## Model & storage
- New `apiKeys` collection (migration **0013**, unique index on the secret hash + index on org).
- Stores only the **SHA-256 hash** of the secret + a display **prefix** (`vsk_1a2b3c4d`), plus `scopes`, `createdBy`, `createdAt`, `lastUsedAt`, `expiresAt?`, `revoked`. The plaintext secret (`vsk_…`) is returned **once** at creation.

## Auth integration (minimal)
- `authenticator` detects a Bearer credential with the `vsk_` prefix and resolves it against `apiKeys` instead of parsing a JWT, then loads the key's **owning user** into the context — so every existing `user.HasRoleFor(org, role)` check works unchanged.
- `lastUsedAt` stamped best-effort (logged, never blocks the request).

## Authorization — deny by default
- API keys are accepted **only** on an explicit allowlist of integrator/managed-org endpoints. Everything else (auth, password, OAuth, team/invites, tickets, checkout, **and API-key management itself**) rejects API-key auth → keys can never touch account-sensitive surface.
- Within the allowlist, each route requires a scope the key must hold:

| scope | grants |
|---|---|
| `quota:read` | GET integrator quota/usage |
| `managed:read` | list managed orgs |
| `managed:write` | create managed orgs |
| `voting:write` | create/publish processes, censuses, bundles |
| `members:write` | manage members & groups |

## Endpoints (admin only)
- `POST /organizations/{address}/apikeys` → create (returns the secret once)
- `GET /organizations/{address}/apikeys` → list (metadata only)
- `DELETE /organizations/{address}/apikeys/{keyID}` → revoke

## Tests (all green locally, Docker-backed)
- `api`: key gen/hash, scope validation, route-allowlist mapping, and a full lifecycle integration test — create → use within scope → **403** for missing scope → **403** for non-allowlisted endpoint → revoke → **401**.
- `db`: storage incl. hash lookup ignoring revoked/expired keys, list, touch, revoke.
- `go build ./...`, `go vet`, swagger regenerated (additive only).

## Design notes for review
- **A key acts as its creating admin user.** This reuses the role model exactly; a key inherits that user's org memberships. If the creator is removed, the key fails closed. Tightening a key to *only* its owning integrator org + that org's managed orgs (rather than any org the user administers) is a natural follow-up — flagging in case you'd prefer that boundary in v1.
- Scope taxonomy is intentionally small/extensible; the allowlist is a single central map keyed by `"<METHOD> <route pattern>"`.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -319,6 +319,9 @@ func (a *API) initRouter() http.Handler {
 		handle(r, http.MethodPost, managedOrganizationsEndpoint, a.createManagedOrganizationHandler)
 		handle(r, http.MethodGet, managedOrganizationsEndpoint, a.managedOrganizationsHandler)
 		handle(r, http.MethodGet, integratorEndpoint, a.integratorInfoHandler)
+		handle(r, http.MethodPost, organizationAPIKeysEndpoint, a.createAPIKeyHandler)
+		handle(r, http.MethodGet, organizationAPIKeysEndpoint, a.apiKeysHandler)
+		handle(r, http.MethodDelete, organizationAPIKeyEndpoint, a.revokeAPIKeyHandler)
 		handle(r, http.MethodPost, subscriptionsCheckout, a.stripeHandlers.CreateSubscriptionCheckout)
 		handle(r, http.MethodGet, subscriptionsCheckoutSession, a.stripeHandlers.GetCheckoutSession)
 		handle(r, http.MethodGet, subscriptionsPortal, func(w http.ResponseWriter, r *http.Request) {
--- a/api/apicommon/apikeys.go
+++ b/api/apicommon/apikeys.go
@@ -1,0 +1,61 @@
+package apicommon
+
+import (
+	"time"
+
+	"github.com/vocdoni/saas-backend/db"
+)
+
+// CreateAPIKeyRequest is the body of POST /organizations/{address}/apikeys.
+type CreateAPIKeyRequest struct {
+	// Human-readable label to identify the key.
+	Label string `json:"label"`
+	// Scopes the key is allowed to use (must be a subset of the assignable scopes).
+	Scopes []string `json:"scopes"`
+	// Optional expiration; when set, the key stops working after this time.
+	ExpiresAt *time.Time `json:"expiresAt,omitempty"`
+}
+
+// APIKeyInfo is the public (no-secret) representation of an API key.
+type APIKeyInfo struct {
+	ID         string     `json:"id"`
+	Label      string     `json:"label"`
+	Prefix     string     `json:"prefix"`
+	Scopes     []string   `json:"scopes"`
+	CreatedBy  string     `json:"createdBy"`
+	CreatedAt  time.Time  `json:"createdAt"`
+	LastUsedAt *time.Time `json:"lastUsedAt,omitempty"`
+	ExpiresAt  *time.Time `json:"expiresAt,omitempty"`
+	Revoked    bool       `json:"revoked"`
+}
+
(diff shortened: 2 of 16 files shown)


lucasmenendez (changes requested):
The CI does not pass

altergui:
Pushed a fix for the **lint** failures (those were real):
- `generateAPIKey` returned 4 values → now returns a small struct (revive `function-result-limit`, max 3);
- split an over-long line in the db API-key test (`lll`, max 130).

Verified locally with `golangci-lint run` (v2.6.x) — `lll`/`revive` clean on the API-key files — plus `go build ./...` and the API-key db + api tests pass.

## The `test` failure is the flaky email test, unrelated to this PR
The `test` job failed only here:
```
--- FAIL: TestRelayVote (14.47s)
    api_test.go: failed to receive email after 10 attempts
    got: e"EOF"   (waitForEmail → testMailService)
```
That's the known flaky CSP-2FA email test (the same `waitForEmail`/MailHog path #534 addressed), not anything this PR touches:
- This PR only adds API keys (db model + storage, the API-key branch in `authenticator`, CRUD handlers, routes, migration `0013`). It changes **nothing** in `notifications/`, `csp/`, `process`, or the mail test harness.
- The auth change is inert for JWT requests: the API-key branch only triggers on a `vsk_` Bearer prefix and otherwise falls through to the unchanged JWT path. `TestRelayVote` uses JWT auth and failed at **email delivery**, well after auth — every step this PR could affect passed.

So this looks like a flaky failure in that email path rather than a regression here. Could we **re-run the `test` job**? Happy to dig in if it reproduces.

emmdim:
### `downAPIKeys` should not `Drop` the collection

`migrations/0013_api_keys.go` reverses with:

```go
func downAPIKeys(ctx context.Context, database *mongo.Database) error {
	return database.Collection("apiKeys").Drop(ctx)
}
```

This breaks the repo's own established migration policy. Migration 0001 (`downInitialCollections`) sets it explicitly:

> Strictly speaking, this down func would Drop all created collections, but that's too risky/destructive. So we do nothing here. (the up func is idempotent anyway)

and 0006/0007 repeat it: `// we do not remove anything to avoid data loss`.

`Drop` *is* the mechanically-correct inverse of `CreateCollection` (and it's idempotent — the driver no-ops on a missing collection), so as pure "undo" it isn't wrong. But "correct inverse" and "safe rollback" diverge for `CREATE COLLECTION`: the only faithful undo is deleting the collection, i.e. data loss. `apiKeys` is the worst place to do that — a rollback silently wipes every live integrator credential.

Suggested fix, matching 0001/0006/0007:

```go
func downAPIKeys(context.Context, *mongo.Database) error {
	// Dropping apiKeys would destroy live credentials; matching the repo policy (see 0001/0006/0007)
	// we do nothing. The up func is idempotent.
	return nil
}
```

If the down should undo *something*, the defensible middle ground is dropping only the indexes (`Indexes().DropAll`), never the collection.

emmdim (approved):
Focuse mostly on auth middleware. Looks ok

lucasmenendez (approved):

PR #559, opened 2026-06-26
fix(api): allow integrator API keys to delete process drafts (voting:write)

altergui:
## The bug

`DELETE /process/{processId}` (delete a voting-process draft) rejects integrator **API keys** with:

```
HTTP 403 {"error":"API keys are not permitted for this endpoint","code":40157}
```

Reproduced live: an integrator using a scoped API key can **create** (`POST /process`), **publish** (`POST /process/{id}/publish`), and **set status** (`PUT /process/{id}/status`) of processes — all `voting:write` scope — but **cannot DELETE a draft**. Because there is also a 2-draft-per-org cap (`40031 max drafts reached`), an integrator that can't publish (e.g. quota-blocked) ends up with drafts stuck permanently, with no key-only way to remove them.

## Root cause

The API-key auth middleware (`authenticateAPIKey` in `api/middlewares.go`) is deny-by-default: it authorizes each request against `apiKeyAllowlist` in `api/apikeys.go`, a map of `"<METHOD> <route>" -> Scope`. Any endpoint **not** present is rejected with `ErrAPIKeyNotAllowed` (code 40157).

The map included `POST /process`, `POST /process/{id}/publish`, and `PUT /process/{id}/status` (all `voting:write`), but was **missing** `DELETE /process/{processId}` — even though delete is the same class of process write op on the same route constant (`processEndpoint`).

## The fix

Add one allowlist entry:

```go
"DELETE " + processEndpoint: ScopeVotingWrite,
```

mirroring the other process write ops. No handler change is needed:

- `deleteProcessHandler` already authorizes via `user.HasRoleFor(org, Manager/Admin)` — identical to the publish/status handlers.
- An integrator API key resolves to its **creating user**, who is admin of the managed org (set as `org.Creator` → `AdminRole` in `SetOrganization`), so the role check passes exactly as it does for publish/status.

Also updated the delete handler's swagger annotation to note API-key support (mirroring publish/status), regenerated `docs/swagger.yaml`, and added a focused integration test asserting an integrator key **can** delete a managed org's draft (and that a key lacking `voting:write` is rejected).

## Tests

- `go test ./api/ -run 'TestIntegratorAPIKeyDeletesProcessDraft|TestAPIKeysAPI|TestIntegratorAPIKeyPathless|TestAPIKeysRequireIntegrator'` → all pass.
- Verified the new test catches the regression: with the allowlist entry removed it fails with the exact 403/40157 from the bug report; with the fix it passes.
- `gofmt` clean; `golangci-lint` reports no new issues in the changed files.

### DIFF
--- a/api/process.go
+++ b/api/process.go
@@ -481,7 +481,10 @@ func (a *API) organizationListProcessDraftsHandler(w http.ResponseWriter, r *htt
 // deleteProcessHandler godoc
 //
 //	@Summary		Delete a voting process
-//	@Description	Delete a voting process. Requires Manager/Admin role.
+//	@Description	Delete a voting process. Requires Manager/Admin role of the organization that
+//	@Description	owns the process.
+//	@Description
+//	@Description	Also callable with a scoped API key (scope: `voting:write`).
 //	@Tags			process
 //	@Accept			json
 //	@Produce		json
--- a/api/process.go
+++ b/api/process.go
@@ -480,9 +480,10 @@ func (a *API) organizationListProcessDraftsHandler(w http.ResponseWriter, r *htt
 
 // deleteProcessHandler godoc
 //
-//	@Summary		Delete a voting process
-//	@Description	Delete a voting process. Requires Manager/Admin role of the organization that
-//	@Description	owns the process.
+//	@Summary		Delete a draft voting process
+//	@Description	Delete a draft voting process. Only unpublished drafts can be deleted; a process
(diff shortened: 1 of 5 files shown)


api/process.go
emmdim, line 501: This should be renamed in `deleteDraftProcessHandler` and make sure the process has no id (=draft)

For now deleting processes is not supported

emmdim (approved):

Issue #560, opened 2026-06-26
Add Go script to re-point orgs orphaned by a deleted custom plan


emmdim:
### Context
A custom plan was deleted from Stripe and consequently removed from the Mongo `plans` collection (Stripe is the source of truth, synced at startup). Orgs that were subscribed still hold the now-deleted `planID` in `subscription.planID`, so `ms.Plan(planID)` returns `ErrNotFound` and everything going through `fetchOrganizationAndPlan` (`db/organizations.go:490`) — counter increments, quota/permission checks — errors out for them. Seen in dev.

### Goal
A Go script (e.g. `cmd/restorecustomplan/`) that:
1. Takes the **ID of a new plan** as a required argument (an already-created Stripe-synced custom plan, present in `plans`).
2. Finds orgs whose `subscription.planID` no longer resolves to a plan in the `plans` collection (optionally scoped to a specific old plan ID via a flag).
3. Re-points each affected org's `subscription.planID` to the new plan ID.

### Notes
- Validate the new plan ID exists first (`ms.Plan(newID)`); bail if not.
- `SetOrganizationSubscription` (`db/organizations.go:257`) replaces the **whole** subscription doc, so do a read-modify-write (preserve `stripeSubscriptionID`, dates, `active`) — or add a targeted `UpdateOrganizationSubscriptionPlanID` helper.
- Follow existing `cmd/` tool conventions (Viper `VOCDONI_` config, like `cmd/service`); reuse the Mongo storage layer.
- Idempotent + dry-run-first (log intended re-points, apply only behind a flag).

PR #562, opened 2026-06-26
feat(api): meter relayed votes and cap elections by vote allowance (MaxVotes)

altergui:
## What

Meters relayed votes for billing and caps usage via a new plan limit `MaxVotes` — **without blocking voters**. Votes always relay; the integrator is billed for what's cast. Enforcement happens at **election publish**: a group census can't be published when its size could exceed the remaining vote allowance.

This mirrors the SMS/email census-size allowance gate and is **built on #557** — the allowance resolves through `limitsOwner`, so:
- a **managed org** draws on the **integrator's shared vote pool** (summed across all its managed orgs via `SumSentVotesManagedBy`);
- a **standalone org** uses its own counter.

`MaxVotes = 0` means **unlimited** (non-breaking: existing plans are unaffected).

## Changes

- **`db.OrganizationCounters.SentVotes`** + `IncrementOrganizationSentVotesCounter`, `SumSentVotesManagedBy` (shared-pool aggregate, mirrors `SumSentSMSManagedBy`).
- **`db.PlanLimits.MaxVotes`** + API surface (`SubscriptionUsage.SentVotes`, `SubscriptionPlanLimits.MaxVotes`).
- **`subscriptions.OrgCanPublishGroupCensus`** — vote-allowance block alongside the email/sms checks: deny when `memberCount > max(0, MaxVotes − sentVotes)`.
- **`POST /vote`** — increments `SentVotes` on a chain-accepted relay (best-effort, in the worker). No quota block at cast time.
- **`ErrProcessCensusSizeExceedsVoteAllowance`** — code `40162`, HTTP **400**, grouped with the email/sms allowance errors.

## Test

- `TestManagedOrgLimits` / `TestManagedOrgSharedPoolExceeded` extended with the vote pool: managed org draws the integrator allowance; standalone uses its own; pool-exhausted is rejected.
- `TestRelayVote` (Docker): votes relay and `SentVotes` increments.
- `golangci-lint --new-from-rev=origin/main`: 0 issues.

## Follow-up (not in this PR)

- Seed real `maxVotes` values into the default/Stripe plan definitions — required before any cap actually bites (0 = unlimited until then).

### DIFF
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -633,6 +633,9 @@ type SubscriptionPlanLimits struct {
 	// Maximum number of census allowed
 	MaxCensus int `json:"maxCensus"`
 
+	// Maximum number of votes that may be relayed; 0 means unlimited
+	MaxVotes int `json:"maxVotes"`
+
 	// Maximum duration of voting processes in days
 	MaxDuration int `json:"maxDaysDuration"`
 
@@ -763,6 +766,9 @@ type SubscriptionUsage struct {
 	// Number of emails sent
 	SentEmails int `json:"sentEmails"`
 
+	// Number of votes relayed
+	SentVotes int `json:"sentVotes"`
+
 	// Number of sub-organizations created
 	SubOrgs int `json:"subOrgs"`
 
@@ -781,6 +787,7 @@ func SubscriptionUsageFromDB(usage *db.OrganizationCounters) SubscriptionUsage {
 	return SubscriptionUsage{
 		SentSMS:    usage.SentSMS,
 		SentEmails: usage.SentEmails,
+		SentVotes:  usage.SentVotes,
 		SubOrgs:    usage.SubOrgs,
 		Users:      usage.Users,
 		Processes:  usage.Processes,
--- a/api/process_vote.go
+++ b/api/process_vote.go
@@ -103,6 +103,11 @@ func (a *API) relayVoteHandler(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			return nil, err
 		}
+		// meter the relayed vote for billing/quota; best-effort, never fail the relay on a
+		// counter error. The quota itself is enforced at election publish, not here.
+		if err := a.db.IncrementOrganizationSentVotesCounter(process.OrgAddress); err != nil {
+			log.Errorw(err, "failed to increment org sent votes counter")
+		}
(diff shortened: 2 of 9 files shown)


emmdim:
Question on the relay-time check (`OrgCanRelayVote` in `relayVoteHandler`):

Since we decided the vote limit is **not** blocking at process time — overage is allowed and the integrator is just billed extra afterwards — the `SentVotes` counter is really the source of truth, consumed later for billing/reporting. Given that, is checking `MaxVotes` synchronously on the relay path worth its cost?

Two downsides of having the check there:
- **Extra hot-path overhead:** two Mongo reads (org + plan) on every relayed vote, to enforce a limit we never actually block on.
- **Side effect — it hard-gates all voting on a resolvable plan.** `OrgCanRelayVote` returns before the cap check on two paths: `PlanID == ""` → `ErrOrganizationHasNoSubscription` (400), and a failed plan lookup → `ErrPlanNotFound` (404). So an org with no/deleted plan can no longer vote at all. That collides with #560 (orgs orphaned by a deleted custom plan would now 404 on every vote), and breaking voting is the worst place for a billing edge case to surface.

altergui:
**Parked — depends on #557.**

Decision: hold the vote-quota *enforcement* until #557 (`limitsOwner()` / managed-org limits → integrator plan) merges, then rebuild the gate on that shared resolver so **sms / email / votes / census** all flow through one path.

Why: on `main` today the sms/email allowance gate (`OrgCanPublishGroupCensus`) and counters are **per-org** even for managed orgs — there's no integrator indirection (that exists only for the aggregate `MaxManaged*` quotas). Since votes should be *billed to the integrator*, the per-org gate in this PR would be throwaway once #557 introduces the integrator-aware resolver.

Kept here as scaffolding to rebase on top of #557:
- `db.OrganizationCounters.SentVotes` + `IncrementOrganizationSentVotesCounter`
- `db.PlanLimits.MaxVotes` (0 = unlimited) and its API surface (`SubscriptionUsage`, `SubscriptionPlanLimits`)
- vote metering at `POST /vote` (counts chain-accepted relays, never blocks)

To revisit after #557: replace the per-org `NEW_PROCESS` gate with the `limitsOwner()`-resolved quota (integrator plan + aggregate counter for managed orgs, own plan otherwise).

emmdim (approved):

PR #575, opened 2026-07-13
fix: two HIGH-severity auth vulnerabilities (JWT expiry, OTP brute-force)

### LINKED ISSUE #574: Review findings: PR #571 (/processes API) + repo-wide security & health audit
Findings from a code review of PR #571 (`feat: multi-question /processes API`) plus a repo-wide security & health audit. Filed as one tracking issue; individual items (especially the security ones) can be split into their own issues/PRs.

---

### HIGH — JWT `exp` set in nanoseconds, so session tokens effectively never expire
`api/helpers.go:59` — `j.Set(jwt.ExpirationKey, time.Now().Add(jwtExpiration).UnixNano())`. jwx v2's `NumericDate.Accept` treats an `int64` as **Unix seconds** (`time.Unix(x, 0)`), so `UnixNano` (~1.75e18) places expiration ~55 billion years out. `jwt.Validate` therefore never rejects on expiry, defeating the intended 15-day lifetime (`jwtExpiration`, api.go:74). The client-facing `LoginResponse.Expirity` is computed separately and correctly, masking the bug. There is no jti/blacklist revocation, so a leaked/stolen token stays valid essentially forever and cannot be invalidated. **Fix:** pass `time.Now().Add(jwtExpiration)` (a `time.Time`) or `.Unix()` seconds.

### HIGH — Password-reset / account-verification OTP has no guess-attempt limit and only 3 bytes of entropy
`VerificationCodeLength = 3` bytes → 6 hex chars (~16.7M values) (`api/apicommon/const.go:30`). `resetUserPasswordHandler` (`api/users.go:664-728`, compare at :707) and `verifyUserAccountHandler` (`api/users.go:166-175`) compare the submitted code with **no attempt counter and no per-account rate limiting** — only the global `middleware.Throttle(100)` concurrency cap. The `UserVerification.Attempts` field / `VerificationCodeIncrementAttempts` exist but are used only in the *resend* anti-spam path, never on the guess path. Code is valid for the full OTP window (default 15 min). **Failure scenario:** attacker triggers a reset for a victim email, brute-forces the 6-hex code online (unlimited guesses, fresh codes every 60s) → full account takeover. **Fix:** enforce an attempt cap on verify/reset (fail closed like the CSP `MaxChallengeAttempts` path) and/or increase code entropy.

### MEDIUM — Global static password salt shared by all users
`api/api.go:75` — `passwordSalt = "vocdoni365"` salts every user's argon2 login-password hash (`internal.HexHashPassword`). Identical passwords yield identical stored hashes across all accounts, enabling cross-account correlation and a single precomputed dictionary against the whole user base. Argon2id params are strong, but the salt should be per-user and random. (Same constant also salts OAuth signature hashes.)

### MEDIUM — API keys are not scoped to their owning organization on address-in-URL endpoints
`api/middlewares.go:90-123` (`authenticateAPIKey`) enforces only the route scope; it never checks the request's target org against `key.OrgAddress`. Handlers like `POST /census`, `POST /organizations/{address}/members`, groups, etc. authorize via `user.HasRoleFor(orgFromURL)`. A key minted "for org X" can act on any other org where the key's creator holds Admin/Manager — blast radius is the creator's full role set, not the owning org. The path-less integrator endpoints correctly use `key.OrgAddress` (organizations_managed.go:28-30), highlighting the inconsistency. Not privilege *escalation* (bounded by the creator's roles), but a least-privilege violation that makes a leaked key far more damaging than its stated scope.

### LOW — CSP signing lock is check-then-act (TOCTOU), mitigated by the DB layer
`csp/sign.go:63-79` — `isLocked()` then `lock()` on a `sync.Map` non-atomically; two concurrent `Sign` calls for the same user+process can both pass the in-memory guard. Contained in practice: `ConsumeCSPProcess` (db/csp.go:216) runs under `keysLock.Lock()` and rejects a second consume with a different address; same-address reuse yields the same nullifier and is deduped by the chain. Worth tightening to a `LoadOrStore` compare-and-swap.

### LOW — CORS allows all origins with credentials
`api/api.go:238-243` — `AllowedOrigins: ["*"]` with `AllowCredentials: true`. Contradictory and browser-ignored for cookie credentials; impact limited because auth is Bearer-token, but replace the wildcard with an explicit allowlist before any cookie/session flow is added.

**Checked and clean:** authorization on protected handlers (consistent `HasRoleFor`, per-request user re-read, no IDOR found), Stripe webhook signature verification, NoSQL injection (typed filters, no user-controlled `$regex`/`$where`), CSP OTP flow (constant-time compare, `MaxChallengeAttempts=5` fail-closed, TTL/cooldown), secret handling (AES-GCM with AAD), notification queue/worker pool. Note: `HashVerificationCode` in `internal/utils.go:114` has a `sha256.Sum()` misuse but is dead code.

---

### 1. `ProcessCheckHandler` skips the `auth.Verified` check
`csp/handlers/processes.go` — the legacy `BundleCheckHandler` requires `auth.Verified` and treats unverified tokens as "not eligible"; the new `ProcessCheckHandler` only checks the process anchor. A caller who knows a member's auth fields (name/surname/email — obtainable without completing the OTP) can call `auth/0` and use the **unverified** token to read the member's census membership, voting weight, per-question eligibility, and `hasVoted` flags — exactly the data the OTP step gates. `ProcessSignHandler` and `ProcessWeightHandler` both check `Verified`; check is the outlier. One-line fix; worth a test (the existing test only covers the wrong-anchor case).

### 2. `updateVotingProcessHandler` checks `Published` (409) before the role check
`api/processes.go` — any authenticated user can probe an arbitrary process ID and learn whether it exists and is published from the 409-vs-401 distinction. Sibling handlers do role-check before state checks. Minor info leak; reorder the checks.

### 3. Batch-result positional contract is the design's soft spot
`account/batch.go` (`mapBatchResults` → `confirmBatch`) — results map back to inputs purely by position, relying on the node grouping fail-fast output as `submitted… / failed / pending…` in input order. There is a length guard, a pinning unit test, and clear comments, but a same-length reordering upstream would silently persist election id A onto question B. Since `Submitted` items carry the predicted `ProcessID`, this could be hardened by locally predicting each question's process id and verifying it matches before persisting. Follow-up hardening, not a blocker — but vocdoni-node #1420 is unmerged and its response shape could still change.

### 4. Minor
- `enqueueStatusChange` submits the on-chain status tx and only `log.Warnw`s if the local `SetQuestionStatus` write fails — chain/DB status can drift until the planned status syncer exists.
- Weight handlers do `big.NewInt(int64(weight))` on a `uint64` — theoretical overflow; sign rejects zero-weight voters but weight/check return 0 silently (inconsistent).
- A few tests use `c.Assert(len(x) > 0, qt.IsTrue)` instead of `qt.Not(qt.HasLen, 0)`.
- `api/processes_review2_test.go` is a working-artifact filename; rename before merge.

### Merge blockers / deployment constraints (already known, restating for tracking)
- `go.mod` pins `go.vocdoni.io/dvote` to the vocdoni-node #1420 **branch commit**; re-pin to the merge commit after #1420 lands. A quiet `go 1.26.1` toolchain bump rides along.
- The publish path is dead against any node without `POST /chain/transactions/batch` — sequence deployment with the node rollout.
- All publish locking (`orgTxMutex`, `db.keysLock`) is in-process only; the PR reinforces the **single-instance deployment assumption**. Do not scale horizontally without distributed locking.

---

## Repo-wide health takeaways

1. **Quota enforcement is spread across multiple entry points and has already been bypassed once** (36f972a fixed unlimited elections via `/transactions`). A single "authorize NEW_PROCESS" choke-point owning both permission and reservation would prevent the next drift — PR #571 adds a fourth publish path.
2. **`validateAuthRequest` in `csp/handlers/handlers.go:746` is a stub** (`// TODO: Add correct validations`) on the public voter-facing path.
3. **`db/plans.go` swallows Mongo errors** (`errors.New("failed to get plan")`, lines 48/67/91) on the path gating every permission decision — also the only violations of the `fmt.Errorf` + `%w` convention.
4. **Breaking API changes are shipping mid-redesign with no versioning story** — three recent `feat(api)!` commits while `plan/*.md` describes phases through 6, integrators already consuming the API.
5. Smaller items:
   - `processBundleParticipantInfoHandler` (`api/process_bundles.go`) is a routed, swagger-documented endpoint that always returns `null`.
   - Bulk CSV member import silently drops per-member failures (`api/census.go:190`, `db/census_participant.go:276`).
   - `cmd/service/main.go` validates config after connecting to Mongo/remote API rather than before; `mongoURL`/`mongoDB` never checked non-empty.
   - Untested: `notifications/smtp` and `notifications/twilio` senders (the failover/queue around them is tested), and `cmd/cli`.

---
*Generated with [Claude Code](https://claude.com/claude-code) — review of PR #571 and repo audit requested by @emmdim.*

emmdim:
Fixes the two HIGH-severity auth findings from #574. One fix per commit.

## 1. JWT `exp` set in nanoseconds → tokens never expire (`fix(auth)`)
`buildLoginResponse` set the `exp` claim to `time.Now().Add(jwtExpiration).UnixNano()`. jwx v2 interprets an `int64` `exp` as Unix **seconds**, so a nanosecond value (~1.75e18) placed expiration ~55 billion years out and `jwt.Validate` never rejected a token on expiry — defeating the intended 15-day lifetime and leaving leaked/stolen tokens valid essentially forever (there is no jti/blacklist revocation).

**Fix:** pass a `time.Time`, which jwx encodes as a proper `NumericDate` (seconds), and reuse the same value for `LoginResponse.Expirity`.

> Note: existing tokens minted with the bad `exp` remain valid until manually invalidated (rotate the JWT secret if you want to force re-login).

## 2. Verification-code / password-reset OTP brute-force (`fix(users)`)
`resetUserPasswordHandler` and `verifyUserAccountHandler` compared the submitted code with **no attempt counter and no per-account rate limiting** (only the global `Throttle(100)` concurrency cap). With a 3-byte / 6-hex code (~16.7M values) valid for the full 15-min window, an attacker could trigger a reset for a victim email and brute-force the code online → account takeover.

**Fix:** new `VerificationCodeCheckAndAddAttempt` — a single atomic conditional `$inc` that only bumps the attempt counter while it is below `VerificationCodeMaxAttempts`, mirroring the fail-closed `IncrementCSPAuthAttempts` guard. Both handlers now spend one attempt **before** comparing the code, so an exhausted code is locked regardless of the submitted value and concurrent submissions can't race past the cap. Reuses the existing `Attempts` field / `ErrVerificationMaxAttempts`.

Defense-in-depth follow-up (not in this PR): raise `VerificationCodeLength` for more entropy. The attempt cap alone already reduces the brute-force space to ≤2 guesses per issued code.

## Testing
⚠️ **I could not build, lint, or run the test suite** — the Go toolchain is not available in the environment where these changes were authored. Changes were verified by manual review only (type-correctness for jwx v2, line-length under the 130-char `lll` limit, `a.db` is the concrete `*db.MongoStorage`). **Please rely on CI** and a maintainer's local `make test` / `make lint` before merging. Regression tests for both paths (an expired-then-rejected token; a code locked after N wrong guesses) are worth adding — happy to follow up.

Closes #574 (partial — the two HIGH items; the MEDIUM/LOW findings remain tracked there).

---
*Generated with [Claude Code](https://claude.com/claude-code)*

### DIFF
--- a/api/helpers.go
+++ b/api/helpers.go
@@ -56,11 +56,15 @@ func (a *API) buildLoginResponse(id string) (*apicommon.LoginResponse, error) {
 	if err := j.Set("userId", id); err != nil {
 		return nil, err
 	}
-	if err := j.Set(jwt.ExpirationKey, time.Now().Add(jwtExpiration).UnixNano()); err != nil {
+	// pass a time.Time so jwx encodes a proper NumericDate (seconds). An int64 here would be
+	// read as Unix *seconds*, so a nanosecond value would push expiry ~55 billion years out
+	// and jwt.Validate would never reject the token on expiry.
+	expiry := time.Now().Add(jwtExpiration)
+	if err := j.Set(jwt.ExpirationKey, expiry); err != nil {
 		return nil, err
 	}
 	lr := apicommon.LoginResponse{}
-	lr.Expirity = time.Now().Add(jwtExpiration)
+	lr.Expirity = expiry
 	jmap, err := j.AsMap(context.Background())
 	if err != nil {
 		return nil, err
--- a/api/helpers_test.go
+++ b/api/helpers_test.go
@@ -1,0 +1,31 @@
+package api
+
+import (
+	"testing"
+	"time"
+
+	qt "github.com/frankban/quicktest"
+)
+
+// TestBuildLoginResponseExpiry guards the JWT expiry regression: the exp claim must be encoded
+// as seconds (a sane ~15-day expiry), not nanoseconds. jwx v2 reads an int64 exp as seconds, so
+// a nanosecond value would push expiration ~55 billion years out and jwt.Validate would never
+// reject an expired token.
+func TestBuildLoginResponseExpiry(t *testing.T) {
+	c := qt.New(t)
+
+	res, err := testAPI.buildLoginResponse("expiry@test.com")
+	c.Assert(err, qt.IsNil)
+
+	token, err := testAPI.auth.Decode(res.Token)
+	c.Assert(err, qt.IsNil)
+
+	exp := token.Expiration()
+	now := time.Now()
+	// expiry is in the future and close to now+jwtExpiration — not decades/eons away
+	c.Assert(exp.After(now), qt.IsTrue)
+	c.Assert(exp.Before(now.Add(jwtExpiration+time.Hour)), qt.IsTrue,
+		qt.Commentf("exp %s is too far out — likely encoded as nanoseconds", exp))
+	// the response-level expiry mirrors the token claim
+	c.Assert(res.Expirity.Before(now.Add(jwtExpiration+time.Hour)), qt.IsTrue)
+}
--- a/api/users.go
+++ b/api/users.go
@@ -163,6 +163,18 @@ func (a *API) verifyUserAccountHandler(w http.ResponseWriter, r *http.Request) {
 		errors.ErrVerificationCodeExpired.Write(w)
 		return
 	}
+	// bound brute-force: atomically spend one verification attempt, failing closed once the
+	// per-code cap is reached. Done before comparing the code so an exhausted code is locked
+	// regardless of the submitted value (no per-guess unlimited retries).
+	recorded, err := a.db.VerificationCodeCheckAndAddAttempt(user, db.CodeTypeVerifyAccount, apicommon.VerificationCodeMaxAttempts)
+	if err != nil {
+		errors.ErrUnauthorized.Write(w)
+		return
+	}
+	if !recorded {
+		errors.ErrVerificationMaxAttempts.Write(w)
+		return
+	}
 	// check the verification code is correct
 	code, err := internal.OpenToken(userVerification.SealedCode, verification.Email, a.secret)
 	if err != nil {
@@ -696,6 +708,19 @@ func (a *API) resetUserPasswordHandler(w http.ResponseWriter, r *http.Request) {
 	// check the verification code is not expired
 	if userVerification.Expiration.Before(time.Now()) {
 		errors.ErrVerificationCodeExpired.Write(w)
+		return
+	}
+
+	// bound brute-force: atomically spend one attempt, failing closed once the per-code cap is
+	// reached. Done before comparing the code so an exhausted reset code is locked regardless of
+	// the submitted value, closing the online guessing vector against the short OTP.
+	recorded, err := a.db.VerificationCodeCheckAndAddAttempt(user, db.CodeTypePasswordReset, apicommon.VerificationCodeMaxAttempts)
+	if err != nil {
+		errors.ErrUnauthorized.Write(w)
+		return
+	}
+	if !recorded {
+		errors.ErrVerificationMaxAttempts.Write(w)
 		return
 	}
 
--- a/api/users.go
+++ b/api/users.go
@@ -168,6 +168,9 @@ func (a *API) verifyUserAccountHandler(w http.ResponseWriter, r *http.Request) {
 	// regardless of the submitted value (no per-guess unlimited retries).
 	recorded, err := a.db.VerificationCodeCheckAndAddAttempt(user, db.CodeTypeVerifyAccount, apicommon.VerificationCodeMaxAttempts)
 	if err != nil {
+		if err != db.ErrNotFound {
+			log.Warnw("could not record verification attempt", "error", err)
+		}
 		errors.ErrUnauthorized.Write(w)
 		return
 	}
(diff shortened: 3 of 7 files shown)


emmdim:
Thanks @copilot — addressed all six comments in 1512f9b:

**Code quality**
- **Log DB errors before returning 401** (`verifyUserAccountHandler` & `resetUserPasswordHandler`): both now `log.Warnw` non-`ErrNotFound` errors from `VerificationCodeCheckAndAddAttempt`, matching the existing `UserVerificationCode` lookup pattern.
- **Misleading error text**: generalized `ErrVerificationMaxAttempts` from `"verification resend max attempts reached"` to `"verification max attempts reached"` (code `40017` unchanged) since it's now returned on the guess path too.

**Tests**
- **DB helper** (`db/verifications_test.go`): `TestVerificationCodeCheckAndAddAttempt` covers increment-to-cap, fail-closed once reached (no increment past cap), `ErrNotFound` vs cap-reached, and per-code-type isolation.
- **Password-reset lockout** (`api/users_test.go`): `TestResetPasswordBruteForceLockout` exhausts the guess budget with wrong codes (401), asserts lockout (400 max-attempts), that the **correct** code is then refused, and that the password is left unchanged.
- **JWT expiry** (`api/helpers_test.go`): `TestBuildLoginResponseExpiry` decodes the minted token and asserts `exp` is a sane ~15-day value, guarding against the nanosecond encoding that pushed expiry ~55 billion years out.

⚠️ Note: the Go toolchain isn't available in my authoring environment, so I could not compile/run these locally — relying on CI to validate.

lucasmenendez (approved):

PR #583, opened 2026-07-17
feat: stage upgrade

elboletaire:


elboletaire:
The PR title lint shouldn't throw errors when merging to stage or lts, since these merges **should not be squashed** nor rebased, which makes the title lint useless since it won't be part of the commit history.

elboletaire:
set as draft until the lint issues are fixed (see #584)

PR #585, opened 2026-07-17
docs: add Fable 5 codebase review (ANALYSIS.md) and refresh CLAUDE.md

elboletaire:
This is Fable 5 speaking on behalf of elboletaire.

I was asked to do a thorough review of the backend and write it up. This PR is that write-up — it's **documentation only**, no source changes. It adds:

- **`ANALYSIS.md`** — a severity-ranked review of the whole codebase (correctness, security, data-integrity, performance, test coverage, structure).
- **`CLAUDE.md`** touch-ups — corrects the test-infra notes (MailHog + in-process Voconed), documents the swagger CI check, Conventional Commits enforcement, the `plan/` rework, and a communication convention.

## How the review was done

Six parallel full-file passes over the subsystems (api core/auth, api domain handlers, db + migrations, csp + account crypto, stripe/subscriptions/notifications/infra, and test coverage), plus `go build`, `go vet`, and `golangci-lint` run locally. Build, vet, and lint all pass today; the 53 lint issues are almost entirely leftover `fmt.Println` debug prints. The single most consequential finding was re-verified end-to-end by reading the code and tracing every call site.

## What it found

**Critical**
- Changing your account email re-derives — and therefore bricks — the on-chain signing key of every organization you created. Org keys come from `HashRaw(secret + creatorEmail + nonce)`, but `PUT /users/me` rewrites `creator` without touching `org.Address`, and nothing validates that the derived signer still matches the account.

**High**
- `BundleSignHandler` is the only CSP bundle handler missing the token↔bundle binding its three siblings enforce (and its census check is commented out) — cross-census vote authorization within an org.
- `SetOrganization`'s error "rollback" can `DeleteOne` an org during a routine update — a data-loss path.
- `HashSortedFields` doesn't actually hash (`sha256.New().Sum(input)` appends a constant), so member PII is stored recoverably in the login-hash keys.
- `publishCensusHandler` stores the account key as the census root instead of the CSP key every other path uses.
- SMTP `Subject` is written raw → header injection via user-controlled ticket titles / org names.
- Bulk-insert error paths nil-deref and, in a detached goroutine, can crash the process.
- Public `GET /organizations/{address}` leaks the Stripe billing email and usage counters.

**Themes** — email-keyed identity is the root of the critical bug and several fragilities; much of the DB layer assumes a single instance (in-process mutex + read-modify-write, unsafe across replicas); the invitation flow missed the recent OTP-hardening pass; and the billing surface is the least-tested area.

Findings carry `file:line` anchors and are tagged `verified` (re-derived this session) vs `reported` (well-cited but not personally re-traced) — worth confirming the `reported` ones before any large refactor. Nothing here changes behaviour; it's meant as a starting point for follow-up issues/PRs.

### DIFF
--- a/ANALYSIS.md
+++ b/ANALYSIS.md
@@ -61,12 +61,12 @@ Cross-cutting themes).
 `UserWeightHandler` (`:483`), and `BundleCheckHandler` (`:580`) all enforce
 `bytes.Equal(*bundleID, auth.BundleID)`. The **signing** endpoint — the one that actually mints
 ballot signatures — does not, and the census-participant check is dead/commented (`:436-439`,
-referencing a `checkCensusParticipant` method that does not exist anywhere). Consequence: a member of
-org O who verifies a token for bundle A (census A) can `POST /process/bundle/B/sign` with that token
-and sign a process in bundle B of the same org, **bypassing census-B eligibility**. Server-side
-weight computation prevents weight inflation, but cross-census vote authorization within an org is a
-genuine integrity hole. **Fix:** add the `bytes.Equal(*bundleID, auth.BundleID)` guard the siblings
-have, and restore a real census-participant check.
+referencing a `checkCensusParticipant` method that does not exist anywhere). Consequence: a token
+verified for bundle A (census A) is accepted at `POST /process/bundle/B/sign` for bundle B of the
+same org, so census-B eligibility is never enforced on the signing path. Server-side weight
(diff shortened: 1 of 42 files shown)


elboletaire:
let's close this since it's more than obsolete and we have new (and in theory better) models to do it whenever

@altergui or @lucasmenendez feel free to create a new PR from scratch if you ever got your hands into this

PR #602, opened 2026-07-23
fix(members): take the org name for import emails from the database

elboletaire:
This is Claude Fable 5 speaking on behalf of elboletaire.

## What

The members-import completion email now takes the organization name from the database (`org.Meta["name"]`) instead of relying on a value that was never populated. The org-name field the frontend used to send for imports is no longer needed (and was already ignored by the backend).

## Why

Since cc0e23f removed the old `OrganizationName` population (it was actually `org.Subdomain`/address hex, deemed confusing), two bugs were left behind:

1. **Dangling subject** — the en/es/ca subject templates still rendered `Members import completed for {{.OrganizationName}}` with the field always empty, producing *"Members import completed for "*.
2. **es/ca emails silently dropped** — their HTML bodies referenced `{{.Organization}}`, a field that doesn't exist on the data struct, so template execution errored and the completion email was never sent for those languages. (Verified by restoring the old template and watching the new all-languages test fail.)

## How

- New `Organization.DisplayName()` accessor for the meta name; used by the import email and by the CSP handlers (three copy-pasted `Meta["name"]` lookups deduplicated into `orgNameAndLogo()`).
- All `members_import_done` templates guard the name with `{{if .OrganizationName}}` so orgs without a name get a clean subject/body.
- Fixed `{{.Organization}}` → `{{.OrganizationName}}` in the es/ca HTML bodies and restored the (now correct) name in the en HTML body.
- Tests: the template is now executed in every language against the real data shape — this catches stale template variables like the `{{.Organization}}` one — plus a no-name fallback case.

Side note: an org whose meta name is an empty string now falls back to the "Vocdoni" default in CSP voter emails instead of rendering an empty name.

No API types changed, so no swagger regeneration needed. For the frontend: any org-name field still sent with import requests is ignored and can be removed at leisure.

### DIFF
--- a/api/apicommon/multilingual_test.go
+++ b/api/apicommon/multilingual_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	qt "github.com/frankban/quicktest"
+	"github.com/vocdoni/saas-backend/db"
 )
 
 func TestMultilingualTextUnmarshal(t *testing.T) {
@@ -57,15 +58,20 @@ func TestBuildOrgMeta(t *testing.T) {
 
 	// only name
 	m := BuildOrgMeta(nil, &name, nil, nil, nil)
-	c.Assert(m["name"], qt.DeepEquals, MultilingualText{"default": "Acme"})
+	c.Assert(m["name"], qt.DeepEquals, map[string]string{"default": "Acme"})
 	_, hasLogo := m["logo"]
 	c.Assert(hasLogo, qt.IsFalse)
 
 	// all three fields
 	m = BuildOrgMeta(nil, &name, &logo, &desc, nil)
-	c.Assert(m["name"], qt.DeepEquals, MultilingualText{"default": "Acme"})
-	c.Assert(m["logo"], qt.DeepEquals, MultilingualText{"default": "https://acme.com/logo.png"})
-	c.Assert(m["description"], qt.DeepEquals, MultilingualText{"default": "We make things"})
+	c.Assert(m["name"], qt.DeepEquals, map[string]string{"default": "Acme"})
+	c.Assert(m["logo"], qt.DeepEquals, map[string]string{"default": "https://acme.com/logo.png"})
+	c.Assert(m["description"], qt.DeepEquals, map[string]string{"default": "We make things"})
+
+	// db accessors must resolve the shorthand values before any Mongo round-trip
+	org := &db.Organization{Meta: m}
+	c.Assert(org.DisplayName(), qt.Equals, "Acme")
+	c.Assert(org.LogoURL(), qt.Equals, "https://acme.com/logo.png")
 
 	// explicit meta wins over shorthand
 	explicit := map[string]any{"name": MultilingualText{"default": "Override"}, "extra": "val"}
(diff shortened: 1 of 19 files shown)


emmdim:
### Code review

Found 1 issue:

1. **`DisplayName()` returns `""` for any org whose name was set through the current API — the email will still omit the org name for most real organizations.**

https://github.com/vocdoni/saas-backend/blob/44675ef3748c7e66bdccced5b93707b080b3dbe1/db/types.go#L100-L112

The new method only handles a plain-string `Meta["name"]`:

```go
name, ok := o.Meta["name"].(string)
```

But since #569 the only write path for the name is `apicommon.BuildOrgMeta` (called from the org create/update handlers), which stores it as a `MultilingualText` locale map (`{"default": "Acme"}`) — even when the client sends a plain string, it is normalized to a map before saving:

https://github.com/vocdoni/saas-backend/blob/44675ef3748c7e66bdccced5b93707b080b3dbe1/api/organizations.go#L147-L149
https://github.com/vocdoni/saas-backend/blob/44675ef3748c7e66bdccced5b93707b080b3dbe1/api/organizations.go#L289-L291

Verified empirically with the repo's driver (`mongo-driver v1.14.0`): after a BSON round-trip, that value decodes as `map[string]interface{}`, so the `.(string)` assertion fails and `DisplayName()` returns `""`. Only legacy orgs whose name predates #569 (and was never updated since) still hold a plain string. The same `.(string)`-on-`Meta["name"]` mistake was previously flagged in #376.

Consequences:

- https://github.com/vocdoni/saas-backend/blob/44675ef3748c7e66bdccced5b93707b080b3dbe1/api/org_members.go#L43-L45 — `org` comes from a real Mongo decode, so `OrganizationName` will be empty for post-#569 orgs. The new `{{if .OrganizationName}}` guards then silently drop the name, so the email renders "Members import completed" without it — the PR looks fixed but mostly isn't.
- The new tests don't catch this: they feed hand-built structs straight into `ExecTemplate`, bypassing both `DisplayName()` and the Mongo round-trip. A test that runs a `db.Organization` through `bson.Marshal`/`Unmarshal` before calling `DisplayName()` would.

The multilingual-aware extraction already exists in `apicommon.OrgDisplayName` / `multilingualFromAny`:

https://github.com/vocdoni/saas-backend/blob/44675ef3748c7e66bdccced5b93707b080b3dbe1/api/apicommon/types.go#L81-L91

`db` can't import `apicommon` (import cycle), so the fix is to make `db.DisplayName()` shape-aware and let `apicommon.OrgDisplayName` delegate to it:

```go
func (o *Organization) DisplayName() string {
	if o == nil || o.Meta == nil {
		return ""
	}
	switch n := o.Meta["name"].(type) {
	case string:
		return n
	case map[string]string:
		return n["default"]
	case map[string]any: // shape after a Mongo round-trip
		s, _ := n["default"].(string)
		return s
	}
	return ""
}
```

Related (pre-existing, worth fixing in the same pass): `orgNameAndLogo` has the identical problem for the **logo** lookup (`Meta["logo"].(string)`) — #569 stores the logo as a locale map too, so post-#569 orgs also fall back to `DefaultOrgLogo`:

https://github.com/vocdoni/saas-backend/blob/44675ef3748c7e66bdccced5b93707b080b3dbe1/csp/handlers/processes.go#L422-L432

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

emmdim (approved):

PR #608, opened 2026-07-28
fix(members): make CSP login robust to whitespace and casing

elboletaire:
This is Claude Opus 5 speaking on behalf of elboletaire.

Makes CSP login robust to the two things that were silently locking voters out of their own elections: **surrounding whitespace** in stored member data, and **letter casing**. Both had the same root cause and the same fix shape, so they land together.

## The bug

CSP login matches a **byte-exact** hash of the census auth fields (`internal.HashSortedFields`). Both sides of that comparison normalized *only* email:

- **Write** — `prepareOrgMember` normalized `Email` and canonicalized `BirthDate`. `Name`, `Surname`, `MemberNumber` and `NationalID` were stored verbatim.
- **Read** — `authFirstStep` built the input member from the request, normalizing `Email` only.

So a member imported with a trailing space in their name could never authenticate. The voter types the clean value, the recomputed hash differs, and the CSP answers `40029 census participant not found` — indistinguishable, from the voter's side, from not being in the census at all. Spreadsheet and CSV imports are exactly where stray whitespace comes from.

The same byte-exactness made login case-sensitive: a member stored as `John Doe` could not log in by typing `john doe`.

And a third instance of the same asymmetry: stored birthdates are always canonicalized to `YYYY-MM-DD`, but the login request was compared raw, so a voter typing `01/01/1990` failed against a stored `1990-01-01`.

## Impact of the whitespace half

Measured on a restored copy of the LTS production database:

| | |
|---|---|
| Members with whitespace in a login field | **56** of 163,674 |
| Voters actually locked out | **142**, across **14** censuses |
| Censuses 100% locked out | **5** |

In the census that prompted the report, 7 of 15 members ever authenticated — and every one of the whitespace-affected members is in the 8 who never did.

## What changed

**One definition of the canonical stored form.** `(*OrgMember).Normalized()` trims the login fields, lowercases the email and canonicalizes the birthdate. Both sides call it — `prepareOrgMember` for storage, `authFirstStep` for the login request — so they cannot disagree. Two independent normalization paths is the bug class this PR exists to fix, so leaving them duplicated would have invited the same drift back.

**Case folding is deliberately not that.** Member documents keep the casing they were imported with, because that is what gets displayed and exported. Only the values that feed the hash are folded, on both sides. So the seam is:

- `OrgMember.Normalized()` governs what is **stored**;
- `HashAuthTwoFaFields` governs what is **compared**, and folds text through the new `LoginHashValue`.

The phone is the exception: it is already a hash, so it is raw bytes rather than text and folding it would corrupt the value.

**The mirror stays in step.** `migrations.hashMemberFields` is a byte-for-byte copy of `HashAuthTwoFaFields`, kept so the repair can work straight from bson documents. It gains identical folding. If the two ever disagree the repair writes hashes login can never match, so `TestRepairMatchesCanonicalHash` drives a matrix of field combinations through the real repair and asserts every hash it wrote equals the canonical one, then resolves each through the real login lookup.

**Existing data is repaired by `scripts/repairlogins`**, in two phases over one run:

- **Phase A** trims member documents.
- **Phase B** recomputes every census participant's hashes and rewrites the ones that changed, so censuses published before this keep working.

Phase B works one census at a time, resolving members with batched `$in` lookups and grouping recomputed hashes in memory rather than issuing a collision query per row. Any group of two or more rows sharing a recomputed hash is skipped whole and reported, since the unique index from migration 8 would reject the second write. That grouping is also what makes rewriting in place safe without a transaction: the stored hash is derived from the field values, so if one row's new hash equalled another's stored hash, that other row's values would already be folded and its own hash unchanged — putting both in the same group. Rows whose member document is gone cannot be recomputed, so their stored hashes are reserved up front and treated as occupied.

## Why a script and not a migration

`RunMigrationsUp` tracks only the **highest** applied version and skips anything `<=` it. Writing a version into the `lts` ledger would mean any lower-numbered migration this branch later absorbs gets skipped **silently** — no error, no warning. `main` already ships a migration at the next free number that creates collections and a unique index, so this is a live hazard, not a hypothetical one. Keeping the repair out of the ledger leaves that ordering intact.

It is the right shape anyway: once the write path normalizes, these rows can no longer be created, so this is a bounded historical backfill rather than an invariant to re-establish. `scripts/defaultplan` and `scripts/fundaccount` are the existing precedent. The function is idempotent, so running it now and again later — including as a registered migration on another branch — is harmless.

## Deploy order is not optional

**Deploy the binary first, then run the repair.** Repairing first would leave any census created in between holding unfolded hashes, with no second pass to catch them. Between the two steps, members whose hash changes cannot log in.

Every affected census was resolved against the production vochain to size that window. The rows whose hash changes sit in 68 censuses backing 61 on-chain processes:

| | processes | affected voters |
|---|---|---|
| **Ongoing** | **1** | **15** |
| Upcoming / planned | 0 | 0 |
| Ended | 60 | 642 |
| Census never bundled | — | 31 |

One live election, 15 voters — but that is **100% of that election's electorate**, so the repair should follow the deploy immediately rather than wait.

An earlier draft proposed a transitional read path accepting both folded and legacy hashes, to make the order irrelevant. At 15 voters rather than 693 that stopped being worth a permanent extra term in the login hot path plus a follow-up to remove it.

```bash
# 1. Deploy the service.
# 2. Report only. Writes nothing. This is the default.
repairlogins --mongoURL "$VOCDONI_MONGOURL"

# 3. Review the summary, especially participantsSkipped.
# 4. Apply.
repairlogins --mongoURL "$VOCDONI_MONGOURL" --apply

# Optional: one organization at a time.
repairlogins ... --org 0x… --apply
```

The database is resolved exactly as `db.New` resolves it: taken from the `mongoURL` path, with `--mongoDB` (or `VOCDONI_MONGODB`) only needed when the URL carries no database or you want to override it. The same environment that runs the service runs this unchanged.

Also available as a `repairlogins` target in `dev.dockerfile`. It connects to MongoDB directly rather than through `db.New`, so it runs no migrations, does not honour `VOCDONI_MONGO_RESET_DB`, and never logs the credentialed connection URI.

## Verification

Full suite green (**636 tests, 31 packages**); `golangci-lint run` clean on everything this touches.

New coverage: the repair (rehash, whitespace, 2FA hash variants, dry run, collision skip, idempotency, org scoping, missing members), the canonical-hash mirror guard, write-path normalization through both the single and bulk paths, birthdate canonicalization, and an end-to-end HTTP login test asserting mixed-case login works *and* that the stored member still reads back with its original casing.

Against a **pristine restore** of the production database, with the service booted against it:

- Dry run reported 56 members to trim and 696 participant rows to rehash across 69 censuses, 12 skipped for collisions, 45,800 orphans — then confirmed to write nothing.
- An independent `mongosh` model predicted 56 / 693 / 68. The 3-row, 1-census difference is fully accounted for: **71 rows had a stored hash that did not match their current member data**, 10 of which trim+fold alone would not have touched. Netting off collision skips, that is the difference exactly.
- **0 collisions caused by folding**, across all three hash variants. The 12 skipped rows collide on identical member data and did so before this change.
- After `--apply`, replaying the real login flow on the live election's census: exact casing still authenticates, lowercase / uppercase / mixed all authenticate, the whitespace cases still authenticate, unknown data is still rejected.
- All 15 members of that census still hold their **capitalized** names in the database.
- A second `--apply` reported zero changes.

## Notes, not action items

- **Post-nominals still match byte-for-byte.** A member imported with the surname `Doe AM` must type `Doe AM`. This is expected and consistent — the stored value is what identifies them — and is a matter for customer communication rather than code.
- **45,800 participants reference a member document that no longer exists**, 45,789 of them in censuses backing a live bundle. Those voters cannot authenticate today regardless, because the CSP loads the member after matching the hash. Deliberately out of scope; recorded because the number is large and the cause (members deleted while their censuses were left behind) may be worth understanding.
- **71 participant rows had stored hashes out of step with their member data** before this ran. The repair brings them into line; how they drifted may be worth a look.

---

Ports to `main` as an identical cherry-pick — see #609.


elboletaire:
Cross-reference: forward-ported to `main` in #609 as an identical cherry-pick — no adaptation was needed.

db/org_members.go
```diff
@@ -171,6 +172,17 @@ func prepareOrgMember(org *Organization, m *OrgMember, salt string, currentTime
 	}
 	member.OrgAddress = org.Address
 
+	// Trim the auth fields. They feed a byte-exact login hash (see
```
lucasmenendez: I think that this should be in a method of the OrgMember struct that returns a normalized version of the data:

`func (m *OrgMember) Normalized() *OrgMember`

This method should include every required transformation, such as `strings.TrimSpace` or `internal.NormalizeEmail` or `normalizeBirthDateInput`. This last one (`normalizeBirthDateInput`) should also moved to the internal package if it can be used by other packages, or to `db` if not, but it should not be in the `api` package.

lucasmenendez (changes requested):

elboletaire:
This is Claude Opus 5 speaking on behalf of elboletaire. All four review comments were valid and are addressed in the latest push.

**@lucasmenendez — `Normalized()` method.** Agreed, and this was the most valuable comment on the PR: two independent normalization paths is exactly the bug class being fixed, so leaving them duplicated invites the same drift again later.

`(*OrgMember).Normalized()` now lives in `db/types.go` and is the single definition of the canonical form. Both sides go through it — `prepareOrgMember` for storage, `authFirstStep` for the login request — so they cannot disagree. `normalizeBirthDateInput` moved out of the handler package to `internal.NormalizeBirthDate`, next to the `ParseBirthDate` it wraps, along with its test.

`prepareOrgMember` keeps validation, which `Normalized()` deliberately does not do: at login an unparseable value should just fail to match a participant, while at write time it has to be reported. Net effect on `csp/handlers/handlers.go` is 28 lines lighter.

**Copilot — connection URI in logs.** Correct, and it matters here specifically because this command is run by hand against production, so its output lands in terminal scrollback and CI job logs. Now logs the hosts only.

Worth flagging separately: this line was copied from `db/mongo.go:99`, which has the same issue — the service logs the full credentialed URI on every boot. Left alone here to keep the hotfix tight, but it looks worth its own change.

**Copilot — "all-or-nothing" overstates the guarantee.** Correct, that wording implied a transaction and there isn't one (multi-document transactions require a replica set, which this does not assume). The doc comment now states what is actually guaranteed: no write happens until every collision check for that member has passed, and participants are updated before the member.

That ordering is deliberate and is the reason a partial failure is tolerable. A participant left with a stale hash behind an already-trimmed member is what locks a voter out, so the write order makes the leftover state the harmless direction — some participants already carrying the trimmed hash while the member still holds the untrimmed values. The member therefore still matches the whitespace prefilter, so a re-run picks it up, rewrites the same hashes to the same values and finishes the job. No special recovery needed.

**Copilot — "indexed-free".** Typo on my part. Reworded to say plainly that a regex cannot be served by an index, so the scan is a collection scan proportional to the member base.

---

Re-verified after the refactor, not just re-tested: full suite green, and the whole flow replayed against a restored copy of the production database with the service booted against it — every previously locked-out voter authenticates, padded input authenticates, every voter who already worked still does, unknown data still rejected, and the repair reports zero remaining work. Same fixes applied to the `main` PR.

lucasmenendez (approved):

elboletaire:
This is Claude Opus 5 speaking on behalf of elboletaire. Pushed: CSP login is now **case-insensitive**, folding the previously-documented follow-up into this PR.

## What changed

Case folding is deliberately **not** the normalization pattern used for whitespace, email and birthdate. Member documents keep the casing they were imported with — that is what gets displayed and exported. Only the values that feed the hash are lowercased, on both sides of the comparison.

So the split is:

- `OrgMember.Normalized()` — governs what is **stored**. Untouched.
- `HashAuthTwoFaFields` — governs what is **compared**. Folds text values through the new `LoginHashValue`.

The phone is the exception: it is already a hash, so it is raw bytes rather than text and folding it would corrupt the value.

`migrations.hashMemberFields` is a byte-for-byte mirror of that function, kept so the repair can work straight from bson documents. It gains identical folding. If the two ever disagree the repair writes hashes login can never match, so `TestRepairMatchesCanonicalHash` drives a matrix of field combinations through the real repair and asserts every hash it wrote equals the canonical one, then resolves each through the real login lookup.

The repair gains a second phase, and is renamed `trimmembers` → **`repairlogins`** to match:

- **Phase A** trims member documents, as before.
- **Phase B** recomputes every census participant's hashes and rewrites the ones that changed, so censuses published before folding keep working.

Phase B works one census at a time, resolving members with batched `$in` lookups and grouping recomputed hashes in memory, rather than a collision query per row. Any group of two or more rows sharing a recomputed hash is skipped whole, since the unique index from migration 8 would reject the second write. That grouping is also what makes rewriting in place safe without a transaction: the stored hash is derived from the field values, so if one row's new hash equalled another's stored hash, that other row's values would already be folded and its own hash unchanged — putting both in the same group. Rows whose member document is gone cannot be recomputed, so their stored hashes are reserved up front and treated as occupied.

## Deploy order is not optional

**Deploy the binary first, then run the repair.** Repairing first would leave any census created in between holding unfolded hashes, with no second pass to catch them. Between the two steps, members whose hash changes cannot log in.

I checked how big that window really is by resolving every affected census against the production vochain. The 693 rows whose hash changes sit in 68 censuses backing 61 on-chain processes:

| | processes | affected voters |
|---|---|---|
| **Ongoing** | **1** | **15** |
| Upcoming / planned | 0 | 0 |
| Ended | 60 | 642 |
| Census never bundled | — | 31 |

So one live election, 15 voters — but that is **100% of that election's electorate**, so the repair should follow the deploy immediately rather than wait.

An earlier draft proposed a transitional read path accepting both folded and legacy hashes to make the order irrelevant. At 15 voters rather than 693 that stopped being worth a permanent extra term in the login hot path plus a follow-up to remove it.

## Verification

Full suite green (**636 tests, 31 packages**); `golangci-lint run` clean.

Against a **pristine restore** of the production database, with the service booted against it afterwards:

- Dry run reported 56 members to trim and 696 participant rows to rehash across 69 censuses, 12 skipped for collisions, 45,800 orphans.
- An independent `mongosh` model predicted 56 / 693 / 68. The 3-row, 1-census difference is fully accounted for: **71 rows have a stored hash that does not match their current member data** (members edited after their census was published), 10 of which trim+fold alone would not have touched. Netting off collision skips, that is the difference. The repair fixes those too, which is correct — participant hashes are already treated as derived from member data by `updateCensusParticipantsForMember`.
- **0 collisions caused by folding**, across all three hash variants. The 12 skipped rows collide on identical member data and did so before this change.
- After `--apply`, replaying the real login flow on the live election's census: exact casing still authenticates, lowercase and uppercase and mixed case all authenticate, the whitespace cases from the first commit still authenticate, unknown data is still rejected.
- All 15 members of that census still hold their **capitalized** names in the database.
- A second `--apply` reported zero changes.

## Runbook

```bash
# 1. Deploy the service.
# 2. Report only. Writes nothing. This is the default.
repairlogins --mongoURL "$VOCDONI_MONGOURL" --mongoDB "$VOCDONI_MONGODB"

# 3. Review the summary, especially participantsSkipped.
# 4. Apply.
repairlogins --mongoURL "$VOCDONI_MONGOURL" --mongoDB "$VOCDONI_MONGODB" --apply

# Optional: one organization at a time.
repairlogins ... --org 0x… --apply
```

## Follow-ups

- ~~CSP login is case-sensitive~~ — fixed here.
- **Post-nominals still must match byte-for-byte.** With a stored surname `Doe AM`, typing `Doe` fails. Unchanged by this PR.
- **45,800 participants reference a member document that no longer exists**, 45,789 in censuses backing a live bundle. Those voters cannot authenticate today regardless, since the CSP loads the member after matching the hash. Out of scope here; noted because the number is large.
- **71 participant rows had stored hashes out of step with their member data** before this ran. The repair brings them into line, but it is worth understanding how they drifted.

elboletaire:
This is Claude Opus 5 speaking on behalf of elboletaire. Two follow-ups pushed:

- **`LoginHashValue` / `foldLoginValue` removed.** They were one-line wrappers around `strings.ToLower` with no behaviour of their own — two differently named aliases for the same stdlib call across two packages, one of them needlessly exported. `strings.ToLower` is now called directly and the reasoning lives in the `HashAuthTwoFaFields` doc comment where it belongs. Net −15 lines.
- **`repairlogins` no longer requires `--mongoDB`** when the connection string already carries the database. It now resolves it the same way `db.New` always has: explicit wins, otherwise the URL path, error only when neither is present. Requiring it twice was an inconsistency I introduced; the service never needed it. Runbook updated.

Issue #597, opened 2026-07-22
Flaky test: TestBlindsaltedKey intermittently fails with "invalid length, need 32 bytes"


lucasmenendez:
## Summary

`TestBlindsaltedKey` (`csp/signers/saltedkey/saltedkey_test.go`) fails intermittently (~1 run in ~128) with:

```
--- FAIL: TestBlindsaltedKey (0.00s)
    e"k error: invalid length, need 32 bytes"
    e"mBlinded error: invalid length, need 32 bytes"
```

Because CI runs `go test -failfast`, a single miss aborts the **entire** test job, so unrelated PRs get red CI (e.g. it tripped on #596, which touches only `api/`/`db/`).

## Root cause

The test feeds random big-int scalars to the blind-signature library, which requires **exactly 32-byte** inputs:

```go
k, signerR, err := blind.NewRequestParameters()            // k is a *big.Int (~256-bit random scalar)
msgBlinded, userSecretData, err := blind.Blind(..., signerR)
blindedSignature, err := sk.SignBlind(salt, msgBlinded.Bytes(), k)
```

`big.Int.Bytes()` (and the internal conversion of `k`) returns the **minimal** big-endian encoding with no leading zeros. A uniformly random 256-bit value has its top byte `0x00` ~1/256 of the time → the encoding is only 31 bytes → the library rejects it with `invalid length, need 32 bytes`. Two values are length-checked per run (`k` and `mBlinded`), so it fails ~1/128 runs.

## Reproduction

```
go test -count=500 -run TestBlindsaltedKey ./csp/signers/saltedkey/
```

surfaces several failures (both `k error` and `mBlinded error` variants).

## Suggested fix

Pad to a fixed 32-byte width instead of the minimal encoding — e.g. `msgBlinded.FillBytes(make([]byte, 32))` rather than `msgBlinded.Bytes()`, and the same for the secret `k` where it is serialized (inside `SignBlind`). That makes the length deterministic regardless of leading zeros.

## Notes

- Pre-existing on `main`; not introduced by any recent PR.
- Worth auditing whether production blind-signing paths serialize these values the same way (if so, the same latent length bug could surface at runtime, not just in tests).

PR #617, opened 2026-07-29
feat(csp): anonymous voting via blind signatures on /processes

lucasmenendez:
## What

Adds **anonymous (unlinkable) voting** to the `/processes` generation: a voter authenticates to the CSP, then receives a blind ECDSA signature over a ballot bundle the CSP never reads. The backend signs and relays the vote without ever learning the voter's on-chain address, so there is no record linking a member to how they voted.

This is a **complete rework** of the earlier branch to mirror [vocdoni-node #1425](https://github.com/vocdoni/vocdoni-node/pull/1425), which reworked the CSP salt derivation almost entirely. The saas-backend's CSP signatures now verify against the reworked chain.

> ⚠️ **dvote is pinned to the vocdoni-node #1425 branch** (`8dd52744108d`). Re-pin to the merged tag before merging this.

## Voter flow

One `/processes` election is one on-chain question (its `UpstreamID`), so the flow is per-question:

1. **`POST /processes/{id}/sign/anonymous/prepare`** — the auth challenge (email/SMS) is resolved; the CSP returns `R` (the blind-signing public parameter) and a **weight attestation**.
2. The voter blinds the CA bundle `{processId, address, voteWeight}` against `R` (client-side; the CSP never sees it).
3. **`POST /processes/{id}/sign/anonymous`** — the CSP blind-signs the blinded scalar and **atomically consumes the single-use session**, recording **no address**.
4. The voter unblinds and casts an `ECDSA_BLIND_PIDSALTED` proof on chain, or relays it through `POST /votes`.

## Fork-aware salt derivation (#1425)

The CSP salt, applied to the census-root key, derives the per-election signing keypair. This is the core of the rework:

|  | Legacy | Post-fork (#1425) |
|---|---|---|
| **salt** | `processID[:20]` | `keccak256(processId ‖ weight[32-BE])[:20]` |
| **scope** | one salt per org per chain | one salt per election |
| **weight** | unbound | hashed into the salt |
| **replay** | blind sig replayable across an org's elections | bound to one election |

The variant is chosen per chain by `genesis.CSPSaltedProofV2Active(chainID, electionStartTime)`, keyed on the election's immutable start time. The CSP carries its `ChainID` (wired in `cmd/service`); an empty or unknown chainID keeps the legacy derivation, matching the chain. Applied to **both** the plain `/sign` (`ECDSA_PIDSALTED`) and the blind path.

## Weight attestation

Because the CSP blind-signs a bundle it cannot read, a voter could inflate the bundle weight. Post-fork the weight is hashed into the salt, so **the chain itself rejects tampering**. As a backstop, `PrepareBlindSign` additionally issues an ECDSA attestation over `(salt, weight)` that the relay (`checkBlindVoteWeight`, resolving the election's start time via `electionStartTime`) verifies before forwarding a blind+weighted vote. The attestation is a backstop — the fork is the boundary.

## Single-use blind sessions

The ephemeral secret `k` lives in Mongo (`cspBlindSessions`, TTL-indexed), not an in-process map: prepare and sign need not land on the same replica, and abandoned attempts drain. Consumption is an **atomic fetch-and-delete**, so concurrent completions yield at most one signature — signing two messages under one `k` would leak the salted private key (`d = (s₁−s₂)/(m₁−m₂)`). `k` is drawn with rejection sampling to a full 32 bytes (regression coverage for #597).

## Tests

- `TestElectionSaltForkAware` — pins the legacy vs. forked derivation across chains.
- `TestBlindSignRoundTrip` / `TestBlindSignRoundTripForkActive` — full blind-sign round trip under each derivation.
- `TestBlindSignSessionIsSingleUse` / `TestBlindSignConcurrentCompletion` — the no-double-sign property.
- `TestCheckBlindVoteWeight` — the relay's weight backstop.
- `TestProcessCSPAnonymousVoting` — casts a real blind-signed vote on the Voconed chain and asserts the CSP recorded no link to the voter's address.

`make lint`, `./scripts/check-qt-patterns.sh`, `make swagger` (in sync), and `make test` are green.


elboletaire (changes requested):

## Overview

Adds a blind-signature signing path to `/processes` so the CSP can authorize a voter over a ballot bundle it cannot read, removing the member → address → ballot link that `ConsumeCSPProcess` currently persists. Two new endpoints, a Mongo-backed session store with a TTL index (migration 0018), a weight attestation checked at the vote relay, plus the `BundleID` → `ScopeID` rename, dead-code removal, and the publish-claim fix.

The engineering is genuinely strong. The comments explain *why* rather than *what*, the crypto reasoning is correct, and `TestProcessCSPAnonymousVoting` casting a real vote on Voconed is the right level of proof. Verified locally: `go build ./...` and `go vet` clean, swagger regenerated, `blind.Point.Bytes()` is fixed-width 33 so the `HasLen, 33` assertion is sound, and `SaltECDSAPubKey`'s mutation is a field reassignment (not in-place `big.Int` mutation), so the `rootPubCopy := *rootPub` shallow copy really is sufficient — that comment is accurate.

Three blocking issues, and one comment that needs rewording because it documents a mechanism that isn't reachable.

---

## 🔴 Blocking

### 1. A voter can cast two counted votes by using both flows

`ConsumeCSPProcessBlind` closes the anonymous budget (`maxOverwrites = 0`) but leaves the **plain** budget open. `db/const.go:27` sets `MaxVoteOverwritesPerProcess = 10`, and `IsCSPProcessConsumed` still asks `TimesVoted > 10`. So after an anonymous signature (`TimesVoted = 1`, `UsedAddress = nil`):

- `ProcessSignHandler` → `authorizeSign` (`csp/handlers/processes.go:356-409`, which performs no consumption check) → `csp.Sign` → `prepareSaltedKeySigner` → `checkProcessConsumed(…, IsCSPProcessConsumed)` → `1 > 10` is false → **proceeds**.
- `finishSaltedKeySigner` re-checks the same predicate, then `markCSPProcessConsumed`'s address-equality guard `address != nil && tokenStatus.UsedAddress != nil && !Equals` (`db/csp.go:287`) short-circuits on the nil stored address, and the plain address is written.

Result: one anonymous vote under address A₁ (nullifier `hash(A₁‖pid)`) plus one plain vote under address A₂ (nullifier `hash(A₂‖pid)`). Two distinct nullifiers, two counted voters, one member. This is the same vote-multiplication class finding #1 sets out to close, reached by mixing flows rather than by repeating one.

The exploit works in one order only — plain-first is correctly blocked, since `IsCSPProcessConsumedBlind` uses `maxOverwrites = 0` and both `PrepareBlindSign` and `CompleteBlindSign` check it. That doesn't reduce severity: the attacker picks the order.

`db/csp_blind_test.go:192-194` currently asserts this as intended behaviour:

```go
// one signature does not exhaust the plain overwrite budget...
consumed, err := testDB.IsCSPProcessConsumed(testBlindUserID, blindProcessID)
c.Assert(consumed, qt.IsFalse)
```

**Suggested fix:** make an anonymous consumption terminal for both flows — either write `TimesVoted = MaxVoteOverwritesPerProcess + 1`, or (clearer) add a `Blind bool` to `CSPProcess` that `cspProcessConsumedBeyond` treats as consumed regardless of `maxOverwrites`. Then invert that assertion and add an API-level test that `POST /processes/{id}/sign` returns 401 after a successful anonymous sign.

### 2. `bundleid` → `scopeid` is a BSON field rename with no migration

`db/csp.go` renames the persisted field (`bson:"bundleid"` → `bson:"scopeid"`) and every filter using it. There is no `$rename` migration — `grep -rn bundleid migrations/` is empty — and `cspTokens` has **no index at all**, let alone a TTL (the only TTL indexes in the repo are on `organizationInvites` and the new `cspBlindSessions`), so every historical `CSPAuth` document survives indefinitely under the old field name.

The description says "No migration: this data is not yet in use." That holds for the `/processes` scope, but the deprecated-but-live bundle flow writes this field — its routes are registered at `api/api.go:468-474` — and its handlers were switched to read the new name:

- `csp/handlers/handlers.go:575` — `!auth.Verified || !bytes.Equal(*bundleID, auth.ScopeID)` → legacy bundle voters silently become `notEligible()`.
- `handlers.go:220`, `handlers.go:477` — resend and weight return 401 "token does not belong to the bundle".
- `csp/handlers/processes.go:371` — same for `/processes` sign, for any token issued before deploy.
- `LastCSPAuth` (`db/csp.go:97`) stops matching → resend cooldown silently bypassed for legacy tokens.
- `CountCSPAuthByScope` / `CountCSPAuthVerifiedByScope` (`cmd/cli`) → 0 for every past bundle; stats break silently.
- `DeleteCSPAuthByScope` (`db/csp.go:498`, org teardown) → deletes nothing, permanently orphaning voter PII. That one has a compliance edge.

To be fair on blast radius: affected voters are not locked out permanently. Re-authenticating issues a token carrying `scopeid`, and `cspTokensStatus` is untouched by the rename, so consumption state survives. The lasting damage is the orphaned historical data and mid-election token invalidation at deploy time.

**Suggested fix:** add migration 0019 doing `updateMany({}, {$rename: {"bundleid": "scopeid"}})` on `cspTokens`, or confirm the collection is empty in every environment before merge. The rename itself is a good call — `ScopeID` genuinely describes both uses — the data just needs to move with it.

### 3. Anonymous weighted votes carry a voter-chosen weight

The CSP blind-signs a bundle it cannot read, and the chain takes `Bundle.VoteWeight` at face value for `ECDSA_BLIND(_PIDSALTED)` proofs. `checkBlindVoteWeight` (`api/process_vote.go:325-359`) is the only check, and the code correctly concedes both of its holes: a voter posting directly to a public Vochain gateway skips the relay entirely, and the attestation is "a bearer token, identical for every voter of that weight" (`csp/sign_blind.go:203-207`), so one shared cert lifts every voter on that election to that weight.

I initially filed this as a discussion point because it is honestly documented. On reflection it belongs here: it is the same integrity class as issue 1, it needs no collusion, and there is no server-side mitigation available in this design. The code comments are right; the problem is that shipping the endpoint makes the gap reachable.

**Suggested resolution:** gate the anonymous endpoints on `census.Weighted == false` until the chain can carry the attestation, and return a clear error for weighted censuses. If you'd rather ship as-is, the API description needs to say plainly that weighted + anonymous is unenforced — right now only the Go comments say it, and operators don't read those.

---

## 🟠 A comment that documents an unreachable mechanism

The `ModifiedCount` → `MatchedCount` change is correct and worth keeping — matching the filter *is* winning the claim, and `MatchedCount` cannot report a win that didn't happen. But the justification attached to it does not hold in production:

`ClaimVotingProcessForPublish`'s filter matches on either `publishing $exists: false` (where `$set` adds the field, so it always modifies) or `publishing $lt (now - PublishStaleAfter)` with `PublishStaleAfter = 15 * time.Minute` (`db/voting_processes.go:153`). A stale marker matched by that second branch is at least 15 minutes old and can never byte-equal the new `time.Now()`. `ClearVotingProcessPublishing` `$unset`s rather than rewrites. So `MatchedCount: 1, ModifiedCount: 0` requires a ≥15-minute clock regression, not routine BSON millisecond truncation — and the "wedges a process in production" claim in both the PR body and the new comment at `db/voting_processes.go:183-189` is not reachable.

The measured flakiness is real, though, and the mechanism is worth writing down correctly: `db/voting_processes_test.go:119` sets `PublishStaleAfter = -time.Minute`. With a negative window the cutoff moves into the *future*, so a marker written moments ago matches the stale branch — and there a same-millisecond reclaim genuinely does write identical bytes. That fully explains 11/20 on main.

So: keep the change, reword the comment. As written it will send the next reader hunting for a production bug that isn't there, and it obscures the real one, which is that the test's negative stale window exercises a path production never takes.

`ClaimProcessForPublish`'s comment is accurate as-is.

---

## 🟡 Minor

- **`DeleteCSPBlindSessionsByElection` is dead and untested** — `db/csp_blind.go:141`, zero references anywhere in the tree including `*_test.go`. In a PR whose stated goal includes removing dead code, this stands out. Wire it into process teardown or drop it.
- **`SetCSPBlindSession`'s `ttl` parameter is always `0`** at its single call site (`csp/sign_blind.go:73-75`), so only the `if ttl <= 0` branch is live. Consider dropping it until something varies it.
- **`isLocked`/`lock` is check-then-act, not atomic** (`csp/sign.go:156-165`). Two concurrent requests can both pass `isLocked` before either calls `lock`, so `sign_blind.go:107`'s "serialize against the plain path" overpromises within a replica as well as across them. Blind `k` reuse is independently safe via the atomic `FindOneAndDelete` — which is the property that matters — so this is a comment-accuracy issue, not a correctness one. Worth saying in the comment that the `FindOneAndDelete` is the real guard, so nobody mistakes the lock for it.
- **`migrations/0018:31` uses a bare type assertion** — `err.(mongo.CommandError)`. Project convention is `errors.Is`/`errors.As`; `errors.As` also survives wrapping. Consider naming the `48` too.
- **`downCSPBlindSessions` drops only the TTL index**, leaving `electionid_1`. Harmless and `up` is idempotent, just asymmetric.
- **`ErrRetryBlinding` for oversized input.** A client sending a >32-byte blinded value is told to "blind again and retry", which won't help — that's malformed, not an unlucky draw. The short-value case is handled correctly.
- **Privacy: timing correlation.** `CSPProcess.UsedAt` records when the voter took their anonymous signature, and the vote lands on chain shortly after. At low turnout that's a de-anonymisation channel independent of the address. Worth a note in the endpoint docs, or coarsening `UsedAt` for blind consumptions.

---

## Done well

- **`ConsumeCSPBlindSession` as `FindOneAndDelete` with `tokenR` in the filter** — atomic single-use `k`, and a stale `R` matches nothing rather than burning the session. The eight-racing-consumers test under `-race` is the right test.
- **Rejection sampling in `newBlindRequestParameters`**, bounded with an explicit error on exhaustion. Correctly identifies #597 as a real-vote failure rather than a flaky test.
- **Domain separation + fixed-width weight** in `weightAttestationMessage` — the byte-compatibility-with-`CAbundle` reasoning is right, and 8-byte BE gives one representation per weight.
- **`ProcessSignInfoHandler` skipping the nullifier when `UsedAddress` is empty**, rather than hashing the zero address into a plausible-looking value that belongs to nobody.
- **`cspBlindSessionID` avoiding `append(userID, …)` aliasing**, with a test using a slice with spare capacity — the case where it would actually bite.
- **The end-to-end Voconed test**, which is what makes the wire-format claims (33-byte R in the library's own encoding, 96-byte uncompressed signature) verifiable rather than asserted.

---

Issues 1–3 are contained fixes and everything around them is in good shape. The publish-claim comment just needs rewording — the code is right.

lucasmenendez:
Deprecated, continued on #641

Issue #337, opened 2025-11-28
if org admin deletes a member, that member can't vote anymore in active processes


altergui:
the most extreme case is deleting the whole memberbase, prevents EVERYONE from voting.

a simple solution would be to check, before deleting a member, if that member is in any active process census, return an error.

and also return an error if an admin is trying to delete the whole memberbase while there is ANY active process (using group-based census)

lucasmenendez:
#621 ships a deliberately narrower rule than this issue asks for, so linking the two and keeping this open as the tracking issue for the broader refusal.

**What #621 does** (Decision 1 in its description): deleting or removing a member is refused with 409 when the CSP has **already signed for them** in a process that is still running (a published question in READY/PAUSED; terminal states release them). So a voter who has obtained a signature can no longer be silently stripped mid-election, and `votes ≤ censusSize` holds. Removals of members who have *not* been signed for go through — and now cascade correctly into the census, the eligibility lists and the CSP sessions, which is the sync this issue's title describes.

**What this issue asks beyond that, consciously not implemented there:**

- refusing the removal of *unsigned* members while a process is active (in #621 the memberbase is the source of truth: an admin removing an unsigned member removes their vote, by design);
- refusing a whole-memberbase delete / group delete while any group-census process is active — #621 declined #613's 409 on group deletion and logs-and-proceeds instead (see `DeleteOrganizationMemberGroup`'s doc comment).

If the broader refusal is still wanted, this is the place to argue it; the guard mechanics to build on are `refuseBlockedVoters` (`api/census_guard.go`) and `OngoingQuestionsByCensuses`. Related follow-up in the opposite direction: #630 (narrowing the existing guard to the election actually affected).

PR #621, opened 2026-07-29
feat(census): keep process censuses in sync with the memberbase

### LINKED ISSUE #611: Make `eligibleMemberIds` updatable also in published processes

lucasmenendez:
Closes #611.

The memberbase is the source of truth. A process census is a **snapshot** of it taken through a group, and a question's `eligibleMemberIds` is a subset of that snapshot — so every memberbase change has to take effect on the censuses of ongoing processes, through the CSP.

On `main` almost none of that held:

| Memberbase change | On `main` today |
|---|---|
| Member deleted | Removed from `orgMemberGroups.memberIds` only. `censusParticipants`, `eligibleMemberIds` and CSP tokens survive → the CSP keeps signing for them |
| Member removed from a group | Nothing cascades — the voter votes on |
| Member added to a group | Nothing cascades — the new member cannot vote a live process |
| New member created | Never reaches a live census built from the auto "All members" group |
| Member info updated | Hashes are recomputed from the **request body** while the member doc is written field-by-field, so a partial `PUT` locks the voter out of every census |
| Question `eligibleMemberIds` | Settable only while the process is a draft (#611) |

And even a correct cascade would not have revoked anyone: `ProcessSignHandler` never re-checked census participation, so a token minted while the voter was in the census kept signing forever.

This is a fresh implementation from `main`. #612 and #613 are not reused; only their review history informed the design.

## Decisions

1. **Removals are refused for voters, not silently applied.** A member the CSP has already signed for cannot be removed while an *ongoing* process exists (a published question that is `READY`/`PAUSED`; `ENDED`/`CANCELED`/`RESULTS` release them). Once voting closes deletion works again, so erasure is never permanently blocked. This is what keeps `census.Size` an honest recount and `votes ≤ censusSize` true — up to one window: the guard reads and then writes, and the CSP can sign in between. The window is a single request round trip, closing it would need a lock shared by the api and csp paths, and the cost of leaving it is one stale signature already bounded by the `cspTokensStatus` consumption row. `refuseBlockedVoters` says so in its doc comment rather than implying the invariant is airtight.
2. **Growth is refused when over plan quota, and changes nothing** — quota is checked before any write, on every path including bulk import. It gates *creations*: an edit of an existing member adds nobody to a census, so it is not preflighted (see *Review fixes*).
3. **Pruning `eligibleMemberIds` to empty opens the question to the whole census** (existing encoding kept), so every prune that empties a published question also enqueues a `maxCensusSize` resize.
4. **Additive API only.** No field removals, and no status-code removals bar one: a malformed member id in `DELETE .../members` used to fail the whole request with a 500 and is now ignored like an unknown one. See *Compatibility*.
5. **Member ids from a request body are scoped to the caller's organization before anything acts on them.** The revocation cascade is org-agnostic by design, so scoping is the caller's job (see *Review fixes*).

## Commits

1. `fix(members)` — merge stored login-hash fields on partial member updates, plus an explicit error switch on the stored read (it was swallowed) and a `created` return. Standalone bug fix. Composes with the `Normalized()` trimming from #609.
2. `fix(members)` — the bulk-import job state raced its progress reporter. **This is the pre-existing `-race` failure on `main`** (`TestOrgMembers/AddBulkOrgMembersLineErrors`). `createOrgMemberBulkOperations` now also returns the inserted ids.
3. `feat(db)` — `census_revocation.go`: `VotingProcessesByCensus` (nothing on `main` queried `votingProcesses` by `censusId`), `OngoingQuestionsByCensuses`, `MembersWithUsedCSPProcesses` (one `Distinct`, and total — these ids come from request bodies and `HexBytesFromString` panics on non-hex), and `RevokeMembersFromCensuses`. Migration 0018 adds the three indexes.
4. `fix(processes)` — `publishing` becomes a struct field and draft mutations are refused with 409 while a publish is in flight. Also fixes `ClaimVotingProcessForPublish` to use `MatchedCount`: **the pre-existing flake in `TestVotingProcessStalePublishReclaim`**.
5. `refactor(processes)` — `enqueueSetProcessCensus(targets)` returns an error instead of writing a response, and never enqueues a shrink.
6. `fix(csp)` — sign-time census re-check plus election-state gates. Without the re-check nothing in the cascade actually revokes anyone.
7. `feat(census)` — the removal cascade wired into every member/group path, the pre-write refusal guard, new error **40173** (409, `data.votedMemberIds`), and `DELETE /processes/{processId}/census`. This changes `DeleteAllOrgMembers`' signature, so the GDPR `EraseUser` teardown from #618 is updated too (it ignores the emptied questions: nothing left to resize during an erasure).
8. `feat(census)` — `api/census_propagation.go` and the three addition hooks.
9. `feat(processes)` — the #611 endpoint.
10. `test(census)` — the end-to-end suite, swagger regenerated.

Then, from the review of this PR (see *Review fixes* below):

11. `fix(census)` — scope member ids to their organization before revoking. **Fixes a cross-organization revocation this PR introduced.**
12. `fix(members)` — the census quota preflight refused plain member edits at the plan limit.
13. `docs(census)` — the guard/sign race window and two scale ceilings, comment-only.
14. `docs(members)` — the delete endpoint's id handling and its 409, swagger regenerated.

Then, from comparing this PR against #612 and #613 before closing them (see *Absorbed from #612 and #613*):

15. `fix(processes)` — refuse eligibility narrowing that strips a signed voter. **Fixes a correctness hole this PR introduced.**
16. `fix(csp)` — re-check census participation on the bundle sign path.
17. `fix(processes)` — report the members actually removed from a census, and bound the request.
18. `test(csp)` — pin the election-state gates.
19. `chore(db)` — drop the feature's index migration.

## Absorbed from #612 and #613

#621 reimplemented both from scratch, so before closing them each was compared against this branch axis by axis: routes, error codes, request/response types, DB methods, CSP behaviour, migrations and tests. #621 was a superset on most — but not all, and one of the gaps was a real bug.

**Whole-census narrowing let a signed voter be stripped (commit 15, from #612).** The eligibility guard was keyed off a diff against the stored list, and a question open to the whole census stores an empty one — the encoding for "no restriction". `diffMemberIDs` derives `removed` by walking that list, so it came back empty and the guard returned early. Restricting "everyone" to a named subset therefore answered **200 with `removed: 0`** while taking the vote from everyone left out, including members already holding a signature the chain accepts. That contradicted Decision 1 and the invariant documented on `blockedVoters`. #612 avoided it by asking *who has voted*; adapted here as `SignedVotersForElections` (the candidate-free sibling of `MembersWithUsedCSPProcesses`) plus `refuseVotersLosingEligibility`. Being keyed off the question's own election also makes it **narrower** than the census-wide guard, which retires the over-broad-409 caveat for this endpoint. `TestUpdateQuestionCensusRefusesStrippingASignedVoter` pins it and fails on the previous commit.

**The bundle sign path never re-checked participation (commit 16, from #613).** `BundleSignHandler` carried the check as three commented-out lines, so the deprecated bundle flow lacked what its `/processes` sibling has. Mostly masked — revocation deletes `cspTokens` by `userid` alone, forcing a re-auth — but not for a participant removal that bypasses that path. Restored mirroring `BundleCheckHandler` in the same file, which already splits `ErrNotFound` from a real DB failure rather than answering 404 for both.

**`removed` was the count requested, not deleted (commit 17, from #613).** `revokeWrites` already held the `DeleteResult` and discarded it, so the honest number was free. `removed` also loses its `omitempty` (it was inconsistent with `added` beside it, and with `UpdateQuestionCensusResponse`), the empty-body reply stops answering a DELETE with `{"added":0}`, and the body is now bounded at 1000 ids — #613's reviewer asked for a cap and #621 had none. Individual malformed ids are still not rejected: with a truthful count they report themselves as removing nothing.

**The election-state gates had no test (commit 18, from #613).** `refuseUnvotableElection` and the unpublished-auth refusal are the refusals most likely to turn away a legitimate voter, and nothing exercised them. Verified by disabling each gate and watching the new test fail.

**Declined, deliberately:** #613's 409 `ErrCensusInUseByPublishedProc` on group deletion. #621 logs and proceeds instead — see the doc comment on `DeleteOrgMembers`' sibling `DeleteOrganizationMemberGroup`. Group teardown mid-election stays possible by design; removals are still refused for members already signed for.

## Review fixes

Two real bugs came out of review, both introduced by this PR, both now fixed with a test that fails without the fix.

**Cross-organization revocation (commit 11).** The revocation cascade resolves *which* censuses to touch from the member ids alone, so an id that reaches it unscoped acts on whatever organization owns that member. `DeleteOrgMembers` revoked the raw request ids before its own org-scoped delete, so a manager of org A submitting org B's ids stripped B's members of their census participation, their question eligibility and their CSP sessions — while the delete matched nothing and answered `count: 0`. Worse, the same request was answered **409 with B's member ids in `data.votedMemberIds`**, leaking ids the caller has no access to.

`db.FilterOrgMemberIDs` is the one scoped lookup, applied inside `DeleteOrgMembers` so no future caller can reintroduce the hole, and in the delete handler so the guard sees the same set the write will. The group path had the same shape — made worse by `revokeWrites` deleting `cspTokens` by `userid` alone, unscoped by census or process — so `UpdateOrganizationMemberGroup` and its handler guard now act only on ids the group actually holds. `TestDeleteMembersIsOrgScoped`, `TestDeleteOrgMembersScopesToOrg` and `TestUpdateOrganizationMemberGroupRevokesOnlyGroupMembers` pin it.

**Quota preflight refused edits (commit 12).** `preflightCensusGrowth` ran before create-vs-update was known, and `OrgCanAddCensusParticipants` asks whether *one more* participant fits — so an organization whose auto-group census sat exactly at its plan `MaxCensus` could no longer edit any of its own members: every `PUT` carrying an existing id was answered 40035. Only a creation grows a census, which is why the propagation below the upsert was already create-only; the preflight now follows the same rule, resolved through the org-scoped `OrgMember` lookup so an id naming no member here still counts as a creation and is still refused before anything is written. `TestUpsertMemberEditAtCensusQuota` pins both directions.

**Not fixed, deliberately.** The removal guard refuses when a member has been signed for *any* ongoing question on the census, not only the question being edited. It is the safe direction, and narrowing it is a behavioural change that deserves its own PR.

## Traps worth reviewing

- **The resize gate is keyed off "does the election need more room", not "were members added".** Reopening a restricted question adds nobody by name yet can multiply the electorate, and `ComputeMaxCensusSize` stamped that election at exactly the subset size — zero headroom. `TestUpdateQuestionCensusReopen` fails without it.
- **`cspTokensStatus` is never deleted.** `ConsumeCSPProcess` pins the single address a member was ever signed for; deleting it would let a removed-then-re-added member be signed for a second address → a second nullifier → a double vote the chain accepts. A test asserts the row survives.
- **The guard runs before the write, everywhere.** Refusing after the group document is updated leaves the member out of the group but still in the census — the exact disagreement this PR removes.
- **Propagation happens on create only.** `AddCensusParticipantsByMemberIDs` re-adds anything missing, so running it on an ordinary member edit would silently undo every revocation. `TestUpdatedMemberDoesNotRejoinRevokedCensus` pins it.
- **`publishing` must stay `bson:",omitempty"`** or every draft ever created looks like a crashed publish.
- **A `$pull` on `eligibleMemberIds` needs an explicit `$type: "array"` guard**: a whole-census question holds a nil slice, which marshals to `null`, and `$pull` fails the whole batch on a non-array value. Found by the tests, not by reading.
- **Nothing may hand the revocation cascade an id it has not scoped.** The cascade is deliberately org-agnostic — it derives the censuses from the members — so scoping is the caller's job, and skipping it reaches into another organization. This is what commit 11 fixes; `FilterOrgMemberIDs` sits in `DeleteOrgMembers` rather than only in the handler so the next caller inherits it.

## The ceiling, stated plainly

**None of this recalls a signature already issued.** The chain verifier checks only that the bundle address matches the tx signer, the process id matches, and the signature recovers to the salted CSP key — no timestamp, no nonce, no revocation list. Revocation closes the door on *new* signatures; a voter already holding one still votes, which is precisely why removals are refused rather than pretended. A test asserts this rather than overselling it.

## Review of this PR (2026-07-30)

Peer review by @elboletaire: 18 findings. 13 acted on, 5 answered in the [reply comment](#issuecomment) — `MED 3` (org lock leak) and `INFO 2` (lowercase ids) do not hold, and three more are deliberate decisions already priced in their commit messages.

The three that changed behaviour rather than prose:

20. `fix(processes)` — scope the census DELETE's member ids to their organization. **Fixes a cross-organization CSP session deletion this PR introduced.** `83ded66` put `FilterOrgMemberIDs` inside `DeleteOrgMembers`; `DELETE /processes/{processId}/census` reaches `RevokeMembersFromCensuses` directly and never inherited it. Two of the three revocation writes are census-scoped, which is why it looked safe — the third deletes `cspTokens` by `userid` alone.
21. `fix(processes)` — retry an eligibility resize the first attempt never enqueued. The empty-diff early return fired before the resize was considered, so a queue-full 503 left the stored list and the election permanently disagreeing.
22. `fix(members)` — report what a census change did, and stop over-counting quota. Propagation errors, resize job ids, the group-add preflight, both stale `@Success` annotations, and `errors.Is`.

`MED 2` (a publish claim landing after the handler's guard read was replaced away) is closed in the merge commit rather than a fix commit, since the write it hardens is the same write `main` rewrote — see `SetVotingProcessDraft`.

## Verification

Rebased on `main` (including the GDPR erasure, vote-batch relay and CSP login-normalization work).

```
golangci-lint run                      # 0 issues
./scripts/check-qt-patterns.sh         # clean
go mod tidy                            # no diff
make swagger                           # no drift
go test -count=1 -failfast -timeout 30m ./...   # 734 passed, 33 packages
go test -race -run TestOrgMembers ./db/         # 24 passed; was failing on main
```

Every new test was also run against the tree *without* its fix and fails there: the cross-org one returns 409 and leaks the other organization's member id, the quota one answers 40035 to a plain edit, the narrowing one answers `200 {"eligible":2,"added":2,"removed":0}` while stripping a signed voter, the removed-count one reports an unknown id as removed, and the election-state gates fail once each gate is disabled.

**No migration.** This is a new feature, so it ships without one — `0018_memberbase_census_sync.go` is gone (commit 19), and 0018 on `main` is now #615's unique question-slot index. The cost is stated in that commit: `cspTokens` has no index beyond `_id`, so revocation scans it, and `VotingProcessesByCensus` scans `votingProcesses` because 0017's index leads with `orgAddress`. Neither is a correctness problem, and the third index the migration created was redundant with 0017's `processId` anyway.

Not done: a manual end-to-end run against `docker compose --profile with-vocone`, and `api/docs.md` is untouched — it documents the legacy single-election `/process` API and has no `/processes` section, so `docs/swagger.yaml` is the contract for these endpoints.

## Compatibility

Additive only, with one exception noted below. `saas-integrator-demo` calls `POST`/`GET`/`DELETE .../members` and stays wire-compatible; the new `censusJobIds` fields are optional and the group `PUT` still answers a bare `OK` when there is nothing to report.

The exception, from commit 11: `DELETE /organizations/{orgAddress}/members` used to answer **500** when any submitted id was not a valid ObjectID, failing the whole request. Those ids are now ignored like the unknown ones they are, so `count` reports the members actually deleted and can be lower than the number of ids sent. Swagger says so explicitly, along with the 409 the removal guard can answer, which was undocumented.


elboletaire (commented):
## Peer review findings

### High

- **`api/processes_admin.go` / `removeVotingProcessCensusHandler`; `db/census_revocation.go` / `revokeWrites` — unscoped member IDs reach the revocation cascade.**
  `DELETE /processes/{processId}/census` passes raw `req.MemberIDs` into `refuseBlockedVoters` and `RevokeMembersFromCensuses` without scoping them to `vp.OrgAddress`. Participant deletion and eligibility pruning are census/process-scoped, but CSP auth-session deletion is by `userid` only, so an org manager can submit foreign member IDs and invalidate another org's CSP sessions. Scope request IDs first with `FilterOrgMemberIDs(vp.OrgAddress, req.MemberIDs)` before the guard/write, and add a regression test mirroring the org-scoping delete tests.

### Medium

- **`api/processes_admin.go` / `updateVotingProcessQuestionCensusHandler` — lost resize retry after queue-full.**
  If eligibility is stored but `enqueueSetProcessCensus` fails, retrying the same request sees an empty diff and returns early, so the on-chain maxCensusSize resize is never retried. Avoid diff-only early return for published questions that may need resize, or make failed enqueue retryable/observable with a job id.

- **`api/processes.go` + `db/voting_processes.go` — process edit can clobber a live publish claim.**
  `SetVotingProcess` uses whole-document `ReplaceOne`; a `publishing` marker set after the handler's guard read can be removed by the replacement. Use a conditional/targeted update that preserves live `publishing`, or explicitly document/accept this TOCTOU.

- **`api/processes_admin.go` / `enqueueSetProcessCensus` — per-org lock leak on queue-full.**
  The org lock is acquired before enqueueing and released inside the queued closure. If enqueue fails because the queue is full, the closure never runs and the lock can remain held. Release the lock explicitly on enqueue failure or acquire it inside the closure.

- **`api/org_members.go` / `upsertOrganizationMemberHandler` — propagation errors are dropped.**
  On member creation, `propagateMembersToCensuses` returns both job IDs and errors, but the handler only returns job IDs. A resize/add failure can be hidden behind HTTP 200. Add `Errors` to `UpsertOrgMemberResponse` or return non-2xx for critical propagation/enqueue failures.

### Low

- **`api/org_members.go` / `upsertOrganizationMemberHandler` — preflight TOCTOU.**
  The creation/edit decision is based on a separate pre-read. If the member is concurrently deleted before the upsert, the write can create without growth preflight. Fix cleanly with an atomic DB-layer quota-checked upsert; unconditional preflight conflicts with edit-at-quota behavior.

- **`api/organization_groups.go` — group add quota preflight over-counts.**
  `preflightCensusGrowth(..., len(toUpdate.AddMembers))` counts IDs already in the group/census, causing false quota rejection. Pre-filter to net-new members before quota preflight.

- **`api/org_members.go` / bulk add — bulk quota preflight over-counts.**
  The preflight uses submitted member count, not actually inserted/net-new count, so duplicate/pre-existing rows can falsely fail quota. Preflight on net-new count where possible, or document the conservative behavior.

- **`api/org_members.go` / async bulk add — propagation result is discarded.**
  Async import ignores propagation errors/job IDs after insert. Persist them into the import job status so clients can see census propagation/resize failures.

- **`api/org_members.go` / delete members — resize job ID is discarded.**
  `resizeEmptiedQuestions` may enqueue a resize, but the delete-member response cannot expose that job. Add job ID(s) to the response, matching group/process census behavior.

- **`api/census_propagation.go` — quota check is non-atomic.**
  `OrgCanAddCensusParticipants` runs before the participant write, so concurrent requests can both pass and exceed the limit. If strict quota is required, re-check inside the locked DB write path; otherwise document as soft limit.

- **`db/census_revocation.go` — revocation writes are non-atomic.**
  Participant deletion, eligibility pruning, and CSP session deletion run sequentially. Partial failures can leave recoverable inconsistent state. Document this mode or add a compensating cleanup/sweep.

- **`csp/handlers/processes.go` — sentinel comparison style.**
  `ProcessSignHandler` uses `err == db.ErrNotFound`; use `errors.Is(err, db.ErrNotFound)` for consistency and future wrapping safety.

- **Swagger annotations/docs are stale for member upsert and group PUT.**
  Upsert now returns `UpsertOrgMemberResponse` with `censusJobIds`, but docs still declare `apicommon.OrgMember`; group PUT can return `UpdateOrganizationMemberGroupResponse`, but docs still say plain `string OK`. Update annotations and regenerate swagger.

### Info

- **`api/census_sync_test.go` — misplaced test doc comment.**
  A comment describing `TestProcessCSPRevocation` appears above `TestRemoveProcessCensusContract`. Move it to the intended test.

- **`api/census_guard.go` — implicit lowercase member-ID invariant.**
  The eligibility guard assumes stored/request IDs line up in lowercase. Normalize or document the invariant.

- **`api/apicommon/voting_processes.go` — `added: 0` on DELETE census responses.**
  Shared response type emits `added: 0` for DELETE. Acceptable, but document if desired.

- **Mongo indexes for revocation paths.**
  Revocation can scan `cspTokens` by `userid` and `votingProcesses` by `censusId`. If revocation traffic grows, add indexes on `cspTokens.userid` and `votingProcesses.censusId`.

lucasmenendez:
Thanks — this was a good pass. Merged `main` (#615 landed on the same draft-update path) and worked through all 18. Verified each against the code rather than taking them as read: **13 acted on, 5 answered below.** Every fix has a test that fails on the commit before it.

## Acted on

**HIGH — unscoped ids → cross-org CSP session deletion.** Confirmed, and worse than it reads: I reproduced it. `83ded66` put `FilterOrgMemberIDs` inside `DeleteOrgMembers`, and this endpoint reaches `RevokeMembersFromCensuses` directly, so it never inherited it. Two of the three revocation writes are census-scoped — which is exactly why it looked safe — but step 3 deletes `cspTokens` by `userid` alone. `TestRemoveProcessCensusIsOrgScoped` asserts on an auth token issued *before* the attack, since a fresh auth papers over the deletion; it answers **401** without the fix. Fixed in `b015132`.

I did not narrow `revokeWrites` step 3 as defence-in-depth. The cascade is org-agnostic on purpose — it derives the censuses from the members, which is what makes it usable from the group and erasure paths — and scoping the session delete to the userids step 1 matched would leave a session alive whenever the participant row was already gone, while hiding the next unscoped caller instead of surfacing it. `FilterOrgMemberIDs`' doc comment already states the contract; every caller now honours it.

**MED 1 — lost resize retry.** Confirmed, fixed in `4c05662` — but not the way suggested. The gate compared the new size against what the *stored list* implied the election was sized for, and that proxy is only right until the two disagree, which is precisely the state being recovered from. So the gate is dropped rather than narrowed: `enqueueSetProcessCensus` already reads the election and enqueues only a genuine increase, so the authoritative check was one layer down the whole time. Net less code, at the cost of one chain read on an idempotent replay.

**MED 2 — publish-claim clobber.** Confirmed, and the mechanism is worse than `ReplaceOne` semantics: the handler *explicitly* zeroes `Publishing` before writing, so the struct-field preservation cannot help. The existing test only covers the read-**after**-claim ordering, and the comment at the write reasons only about a *stale* marker — silently wrong for one that goes live after the guard read. Closed in the merge commit: `SetVotingProcessIfUnchanged` becomes `SetVotingProcessDraft`, absorbing `storeUpdatedProcess`, with the `updatedAt` token now optional *inside* the function and the filter also requiring an unpublished draft nobody is publishing. `TestSetVotingProcessDraftRefusesPublish` deliberately sends no token — the guarantee must not depend on a client opting into optimistic concurrency, since `updatedAt` is optional on the wire.

**MED 4 + LOW 5 + LOW 2 + LOW 9 + LOW 8 + INFO 1** — all as described. `UpsertOrgMemberResponse` gains `errors` (both sibling callers already surfaced it); `DeleteMembersResponse` and the group DELETE gain `censusJobIds`; the group-add preflight now counts net of `group.MemberIDs`, mirroring the `removedInGroup` filter directly above it; both stale `@Success` annotations fixed and swagger regenerated — neither type was in `swagger.yaml` at all, so the generated contract was wrong; the four sentinel comparisons this PR added are now `errors.Is`, including the `switch err` that could not have been converted mechanically later; the stray test doc comment moved to the test it describes.

**LOW 1, 3, 6, 7** — comments, as you framed them. LOW 7 in particular was genuinely undocumented (`5d2e32a` documented a different race), so `revokeWrites` now records that the three steps are not transactional, that step 1 is the load-bearing one, and that the residue of a partial failure is cosmetic and idempotently repaired.

## Answered, not changed

**MED 3 — org lock leak. This one is wrong.** `enqueueTx` → `enqueueTxBatch` is all-or-nothing and returns `false` having sent nothing, and the queue-full branch unlocks explicitly:

```go
}}) {
    orgLock.Unlock()                // ← api/processes_admin.go
    if e := a.db.SetJobStatus(jobID, db.JobStatusFailed, nil, "tx queue full"); e != nil {
```

Exactly one of the two `Unlock()` calls ever runs. `processes_publish.go` implements the same discipline with a `lockHeld` flag.

**HIGH, second half — `updateVotingProcessQuestionCensusHandler` has the same gap.** It does not. That handler runs `req.MemberIDs` through `validateCensusParticipants` first, which 400s on any id that is not already a participant of *this* census, and it writes via `SetQuestionEligibleMemberIDs` — it never reaches the cascade, so there is no `cspTokens` delete on that path.

**INFO 2 — lowercase member-id invariant.** Enforced upstream rather than assumed. `voters` come from `cspTokensStatus.userid`, written as `HexBytes.String()` → `hex.EncodeToString`. `allowed` is not raw input: it must survive an exact-match `$in` against stored `participantID` values, themselves `member.ID.Hex()`, or it is a 400 before the guard sees it. The one place that silently drops a mis-cased id is `FilterOrgMemberIDs`, and that fails safe by the documented contract.

**INFO 3 — `added: 0` on DELETE.** Deliberate. The `Removed` doc comment argues for always-present counters precisely so zero is distinguishable from absent; adding `omitempty` to `Added` reintroduces that ambiguity on the other field, and splitting the type churns the API for a cosmetic zero.

**INFO 4 — revocation indexes.** Deliberate, and priced in `aa74686`'s commit message — `cspTokens` scanning by `userid` and `votingProcesses` scanning by `censusId` are both named there, along with why the third index was redundant. The migration was dropped because version 18 was contended by three in-flight branches, not because the cost was unnoticed; `main` now has 0018 from #615. Re-adding it here would re-take a contended number for a feature with no data yet. Correct follow-up is a separate migration once numbering settles.

**LOW 4 — async bulk propagation.** Correctly scoped as a known limitation; the code says why (`CompleteJob` fires before propagation runs). The fix needs a `db.Job` schema field, a new db method — `CompleteJob` `$set`s `errors` wholesale, so a second call clobbers rather than merges — and a response-shape change. Filed as follow-up rather than smuggled in here. The better version if revisited: reorder so `CompleteJob` runs *after* propagation and merges its errors, which also fixes the misleading `completedAt`.

**LOW 6 — quota atomicity.** Pre-existing: `OrgCanAddCensusParticipants` has had this read-then-compare shape since `666aef2d` (Feb 2026), and `api/census.go` and `processes_admin.go` already call it that way on `main`. This PR adds callers, not the race. Documented on `preflightCensusGrowth` as a soft limit rather than changed, since making it hard would put a `subscriptions` dependency in the `db` layer.

## Also in this push

Merged `origin/main`. #615 rewrote `writeQuestions` (per-slot replacement) and added the optional `updatedAt` token to `PUT /processes/{id}` — same handler this branch guards, so three files conflicted. Two auto-merges were load-bearing and checked rather than assumed: #615's `_id,omitempty` lands on `VotingProcessQuestion` (not `VotingProcess`, despite the hunk header) and `SetProcessQuestions` depends on it; and the `publishing` marker survives #615's new writers for free, since the mechanism is `bson:"publishing,omitempty"` in `db/types.go`, not anything in `SetVotingProcess`.

The PR body's "`main` still tops out at 0017" is now stale — updated.

elboletaire (approved):
This is Fable speaking on behalf of elboletaire.

I compared the review findings against the four new commits (and the `main` merge) and verified each fix in the code rather than taking the response comment as read. Approving.

**Verified fixed:**
- **HIGH — unscoped ids reaching the revocation cascade** (`b015132`): `FilterOrgMemberIDs(vp.OrgAddress, …)` now runs before both the blocked-voter guard and `RevokeMembersFromCensuses` in `removeVotingProcessCensusHandler`, matching the delete-members handler. `TestRemoveProcessCensusIsOrgScoped` pins it the right way — asserting on an auth token issued *before* the attack, so a fresh re-auth can't mask the deletion.
- **MED 1 — lost resize retry** (`4c05662`): dropping the stored-list size gate instead of patching it is the better fix — the gate's proxy was wrong exactly in the recovery state, and `enqueueSetProcessCensus` already reads the election and only enqueues genuine increases. The empty-diff early return now applies to drafts only, so a replay after a queue-full 503 does retry the resize.
- **MED 2 — publish-claim clobber** (merge commit): `SetVotingProcessDraft` makes the precondition part of the write (`published: false` + no live `publishing` marker, `updatedAt` token optional), which is the correct shape given the handler's read is stale by write time. `TestSetVotingProcessDraftRefusesPublish` correctly avoids relying on the optional token.
- **MED 4 + LOWs** (`ac5947a`): `UpsertOrgMemberResponse.errors` populated on create; group-add preflight counted net of `group.MemberIDs` with an honest comment about the residual over-count; `censusJobIds` on the delete-members and group-delete responses; `errors.Is` conversions; both stale `@Success` annotations fixed and swagger regenerated; `revokeWrites` non-transactionality documented.

**Rejections independently confirmed:** MED 3 is indeed wrong (the queue-full branch unlocks explicitly before `SetJobStatus`); the question-census endpoint never reaches the cascade (`validateCensusParticipants` 400s any id that isn't already a participant); and the lowercase-id invariant is enforced upstream by the verbatim `$in` match. The deferred items (async bulk propagation, revocation indexes post-0018-contention, quota soft-limit) are reasonably scoped as follow-ups.

Build and the new regression tests are in place. Nice work on the fix-with-failing-test discipline throughout.

emmdim:
Fable 5 here, controlled by emmdim.

Read the full history before this lands — #611, both closed PRs with their review rounds, and the 18-finding pass here. None of the following blocks the merge; all four are about commitments and loose ends that would otherwise evaporate when this PR closes.

**1. The promised follow-up issues were never filed.** The review reply above says the async bulk propagation gap was *"Filed as follow-up rather than smuggled in here"* (LOW 4), and the revocation-index deferral (INFO 4) says *"Correct follow-up is a separate migration once numbering settles."* I searched the tracker: no issue exists for either, nor for the third agreed deferral (narrowing the removal guard to the question being edited, named in the PR body under *Not fixed, deliberately*). As it stands these commitments live only in this comment thread. Please file the three issues before or at merge:
- async bulk-import propagation reporting (needs the `db.Job` schema field + `CompleteJob` merge semantics described in the reply),
- indexes on `cspTokens.userid` and `votingProcesses.censusId` once the 0018 contention settles,
- guard narrowing as its own behavioural PR.

**2. Consider closing #337 — or make it the place where the broader guard is argued.** Open issue #337 asks to refuse deleting members who sit in an active process census, with "delete the whole memberbase mid-election" as its extreme case. This PR ships a deliberately narrower rule: only *signed-for* voters are refused; unsigned members can still be revoked mid-election, and group deletion logs-and-proceeds (the #613 guard was declined on purpose). That is a defensible product call — but it means #337's ask is consciously *not* implemented, and the issue is neither referenced nor closed by this PR. Either close #337 with a pointer to Decision 1 here, or keep it open as the explicit follow-up for the broader refusal. It shouldn't stay open unlinked.

**3. Flagging the CSP token-hygiene debt as a follow-up issue.** #613's description recorded it and it did not survive that PR's closure: verified CSP tokens have no TTL and no expiry index, `VerifiedAt` is written but never read (still true on this branch, `db/csp.go:32`), and `DeleteCSPAuthByBundle` is only reached via `ProcessBundlesByOrg`, so `/processes`-flow tokens survive a full organization teardown. This branch's sign-time re-check narrows the impact (a stale session can no longer sign after revocation), but an abandoned verified session remains resumable indefinitely. Deserves its own issue so it stops living in closed-PR descriptions.

**4. Async bulk import still discards propagation results — is that on purpose?** `api/org_members.go:282` drops the return value of `propagateMembersToCensuses`, while the sync path forty lines up (line 240) surfaces jobs and errors to the caller. The review reply explains the mechanics (`CompleteJob` fires before propagation and a second call would clobber `errors`), so I assume it is deliberate — but unlike every other accepted limitation on this branch, the call site carries no comment saying so. If intentional, a one-liner there plus the follow-up issue from point 1 keeps it from reading as an oversight; if not, it should block the follow-up, not this PR.

lucasmenendez:
All four points acted on — you were right that these commitments lived only in this thread.

**1. Follow-up issues filed:**
- #628 — async bulk-import propagation reporting (the `db.Job` field / `CompleteJob` merge semantics / reorder-after-propagation sketch from the reply above).
- #629 — indexes on `cspTokens.userid` and `votingProcesses.censusId`, once migration numbering settles; also records why the third original index stays dropped.
- #630 — narrowing the removal guard to the election actually affected (the *Not fixed, deliberately* item from the PR body).

**2. #337** stays open, now linked: commented there spelling out what Decision 1 ships (signed-for voters refused, terminal states release) versus what #337 keeps asking beyond it (refusing unsigned members mid-election, refusing memberbase/group teardown while a group-census process is active), and pointed at the guard mechanics a broader refusal would build on. It is now the tracking issue for that product debate.

**3. CSP token hygiene:** filed as #631 — no TTL, `VerifiedAt` write-only, `DeleteCSPAuthByBundle` unreachable for `/processes`-flow tokens — with the note that this branch's sign-time re-check narrows but does not close it.

**4. The async discard is deliberate**, and one correction: the call site did already carry a comment ("a failure here is logged, not reported") — what it lacked was the *why* and a pointer. `f3ae229` extends it to name the CompleteJob clobber mechanics and #628. Nothing behavioural changed.

Also pushed since your comment, from a second review pass (`823572c`, `98ce7e9`, `0aaec02`): removal-side resize failures now surface in the response bodies instead of only the log, a recount failure no longer loses an emptied question's resize, and the re-add-after-revocation flow the `cspTokensStatus` comment promises is now driven end to end by a test — which caught `IsCSPProcessConsumed` erroring on the dangling `UsedToken` the cascade leaves behind, fixed in the same commit.

lucasmenendez:
Thanks — four of the six were real, and two of those were worse than the write-up suggested. Each was verified against the code before acting on it rather than taken as read; every fix has a test that fails on the commit before it.

**Acted on: H1, H2, M1 (which subsumes L2), L1. Answered: M2.**

## Acted on

**H1 — group re-add undoes a revocation.** Confirmed, and the chain is complete: the revocation cascade is census-scoped and never touches `orgMemberGroups`, so a member revoked via `DELETE /processes/{pid}/census` stays in `group.MemberIDs`. A later group `PUT` naming them passed the quota check (`addedNotInGroup` counted them as 0), the group write deduped to a no-op — and propagation still handed the id to `AddCensusParticipantsByMemberIDs`, which re-inserts anyone missing. The member was missing *because* revocation deleted their row, so the revocation was silently undone behind a 200.

Fixed in `5481b3c` by propagating only the members genuinely new to the group. The set is computed once and reused for both the preflight (which already counted net-new) and the propagation, mirroring `removedInGroup` on the removal path. Notably the sibling member-edit path already defends against exactly this (`api/org_members.go`, "propagate on create only", pinned by `TestUpdatedMemberDoesNotRejoinRevokedCensus`) — the group path had the same primitive with none of the guard. `TestGroupReAddDoesNotUndoRevocation` pins both directions in one request: the revoked member stays out, a genuinely-new member still joins.

**H2 — `autoGroupCensuses` swallows lookup errors.** Confirmed. The signature had no error return, so a wrapped infrastructure failure and a legitimate `ErrNotFound` both produced `nil`, distinguished only by a log line. Both consumers read empty as "nothing to do" — `preflightCensusGrowth` returns nil (quota OK), `propagateMembersToCensuses` returns a zero-valued result with no errors — so a transient Mongo blip created the member, consumed no quota, propagated nowhere, and answered **200 with a clean body**. The durable half is the sharper one: memberbase and censuses left permanently out of step, with no job to poll and nothing to retry from.

Fixed in `0bd78f3`: the signature becomes `([]string, error)`, `ErrNotFound` stays the legitimate empty case, anything else is wrapped and returned. Both callers (upsert, bulk import) now abort **before any memberbase write**, so the request fails retryably instead of succeeding dishonestly. That is a deliberate availability-for-consistency trade on a write whose whole job is keeping the two in step. `TestAutoGroupCensusesReportsLookupFailures` pins all three outcomes.

**M1 — resize targets lost after commit — and L2 with it.** Confirmed, with one correction and one aggravation. The correction: the *candidate set* was already captured pre-write, so the loss window is narrower than "no pre-read exists". The aggravation: the comment directly above the post-commit read says a failure past that point "must not discard it or the caller's resize is lost for good" — and the very next line does exactly that, `return 0, nil, err`, discarding both the committed `removed` and the `emptied` list. The mitigation had been applied to the recount below it but not to the read it was written about. Retry cannot reconstruct: the `$pull` has already pruned the names, and a `[]` eligibility list is byte-identical to a question that was always whole-census.

Fixed in `499738e` by deciding emptiness **before** the write instead of hardening the read: a question is emptied iff every member it names is being revoked. `publishedQuestionsNaming` already decoded the questions only to throw away all but the id, so returning them costs nothing, and `emptiedByRevocation` selects from them. No failure after the commit can lose the targets, and on a partial-commit error they are returned rather than nilled.

That deletes the `emptiedQuestions` re-read and its `$in {nil, []}` filter outright — **which is L2**. Whole-census questions never reach the check (the candidate filter excludes them), so the null/missing/empty-array equivalence no longer applies; the reasoning is documented on `publishedQuestionsNaming`, where it now matters. Worth flagging: the test L2 credited with encoding that behaviour was **structurally vacuous** — in all three cases its loop ran over an empty `emptied` list, so the assertions could not fail whatever the filter did. It is replaced with a positive test of the strict-subset boundary. Net −19 lines.

**L1 — "voted" naming.** Agreed, with a nuance: the wire *message* already said "signed for", and the doc comment already said the "voted" wording must never be reported — only the Go identifier and the JSON key still contradicted it. Renamed in `24fc88c` to `ErrCensusMemberAlreadySignedFor` / `data.signedMemberIds`, swagger regenerated. No compatibility path needed: 40173 and the field were introduced by this branch and are absent from `main`, so nothing shipped consumes either name. Marked `!` since it is a wire rename regardless.

## Answered, not changed

**M2 — async import completes before propagation.** All four mechanical points hold, and it is deliberate rather than overlooked — `f3ae229` documents it at the call site with the `CompleteJob` clobber mechanics (it `$set`s `errors` wholesale, so a second write destroys the import errors rather than merging the propagation ones) and a pointer to **#628**, filed for it. The real fix needs a `db.Job` schema field, new merge semantics, and a response-shape change; the better shape, as noted there, is reordering so `CompleteJob` runs *after* propagation — which also fixes the misleading `completedAt`. That is its own PR, not a rider on this one.

## Verification

```
golangci-lint run                 # 0 issues
./scripts/check-qt-patterns.sh    # clean
go mod tidy                       # no diff
make swagger                      # only the L1 rename, no drift
go test -count=1 -timeout 30m ./...   # 17 packages, 0 failures
```

Each new test was run against the tree without its fix: the group one re-adds the revoked member, the auto-group one reports no error on a real failure, and the subset one is the replacement for a test that could not fail at all.

elboletaire (dismissed):

Re-reviewed the full branch (`origin/main...HEAD`, 30 commits) with independent reviewers per area, then re-derived every blocking candidate by hand before reporting. The plan is unusually honest and almost every claim in the PR description holds as written — the findings below are gaps at seams the plan didn't map, not defects in what it did map. Three of them are blocking.

### Blocking

1. **`DeleteAllOrgMembers` bypasses the auto-group census guard and permanently severs propagation.** `db/org_members.go:787` calls the unguarded `deleteAutoMemberGroup`, while `DeleteAutoMemberGroupIfEmpty` (`db/org_members_groups.go:750-753`) explicitly refuses when `len(group.CensusIDs) > 0` — its own comment names the hazard: `EnsureAutoMemberGroup` recreates the group with a fresh ObjectID, orphaning every census that references the old one. `group.CensusIDs` is written in exactly one place (`addOrganizationMemberGroupCensus`, at census creation), so nothing ever relinks. Reachable: `DELETE .../members` with `all: true` passes `refuseBlockedVoters` whenever nobody has been signed for yet. Consequence: the obvious wipe-and-reimport workflow ("my CSV was wrong") leaves the live census permanently empty — `autoGroupCensuses` resolves the recreated group's empty `CensusIDs`, re-imported members never propagate, nobody can vote, and no error surfaces anywhere. Fix is one line: call `DeleteAutoMemberGroupIfEmpty` there. The org-teardown caller (`api/organizations_managed.go:536`) is unaffected since teardown also runs `DeleteAllOrgMemberGroups`.

2. **The removal guard is blind to legacy bundle elections whose censuses the cascade nonetheless revokes.** `CensusesForMembers` (`db/census_revocation.go:48`) distincts over all of `censusParticipants`, which bundle censuses share (their participants are org members — see `csp/handlers/handlers.go:428,608,913`), and `revokeWrites` deletes their rows and recounts `census.Size` down. But `blockedVoters` resolves "ongoing" exclusively through `ms.votingProcesses` (`OngoingQuestionsByCensuses`), so a bundle-backed census yields zero questions and the guard waves the deletion through. A member the CSP already signed for on a bundle election is therefore deletable while their signature still votes — the invariant documented on `blockedVoters` itself ("no signature the chain will accept can belong to someone who is no longer counted") is false for bundle censuses. On top, a delete → re-import (new ObjectID) → re-add gets a fresh `cspTokensStatus` row and hence a second address/nullifier — the double-vote pin is keyed on the member id, so it does not survive recreation (this chain is inferred from reading, not constructed). The PR clearly still maintains the bundle path (it restored its participation re-check), so this needs a decision: extend `OngoingQuestionsByCensuses` to bundle processes, or explicitly scope the invariant claim to `/processes` in the doc comments and accept the gap.

3. **Terminal questions are pruned and fed to the resize job.** `publishedQuestionsNaming` (`db/census_revocation.go:335-339`) has no status filter, while its guard-side sibling `OngoingQuestionsByCensuses` filters to READY/PAUSED. Two consequences. (a) An ENDED/CANCELED question emptied by a removal has its `eligibleMemberIds` silently rewritten, destroying the record of who was eligible in a completed vote. (b) It is returned in `emptied` and reaches `enqueueSetProcessCensus`, where the dead election reads fine, `current` is the zero-headroom subset size, and the post-revoke `census.Size` often exceeds it — so a `SET_PROCESS_CENSUS` tx is submitted for a dead election, the chain refuses, and the job loop's first-error return (`api/processes_admin.go:676-688`) aborts every remaining target in the same job, including live elections' legitimate resizes. Visible only as a failed async job — the silent-divergence class this PR exists to eliminate. Reachable because the guard deliberately releases terminal questions' voters. Fix: mirror the READY/PAUSED filter into `publishedQuestionsNaming`, and decide deliberately whether the `$pull` should spare terminal questions too.

### Non-blocking, worth a follow-up

- **A failed resize enqueue on the revocation paths is not recoverable by replay** — `publishedQuestionsNaming` filters on `eligibleMemberIds $in memberIDs` and the `$pull` consumes its own trigger, so replaying the DELETE re-derives nothing. There *is* a working recovery (`PUT .../questions/{qid}/census` with `[]` — the CAS matches the emptied list and `enqueueSetProcessCensus` re-derives the increase from the chain), and the failure is surfaced in the response `Errors`, so this is a documentation gap: say on the revocation endpoints what the recovery lever is.
- **`revokeWrites` returns `removed: 0` on every error path** (`db/census_revocation.go:420,428`), including step-2/3 failures after step 1 committed N rows; and all five callers of `RevokeMembersFromCensuses` drop `emptied` when `err != nil`, so the stated benefit of computing resize targets before the write is currently unreachable. Both matter only for logs/diagnostics today (callers 500 and discard), but they contradict the "honest counts" intent.
- **The guard reads the optimistically-written stored status** (`api/processes_publish.go:763-768` writes it before `enqueueConfirm` verifies) — a failed status tx leaves a seconds-wide window where stored=ENDED, chain=READY, and a signed voter is deletable. Same accepted-risk class as the documented guard/sign race; deserves a sentence in the doc comment.
- **The 1000-id cap guards only `DELETE /processes/{processId}/census`** — the members DELETE, the group PUT and the eligibility PUT reach the same cascade unbounded; the stated rationale applies verbatim to all of them.
- **`FilterOrgMemberIDs` is case-sensitive** (`db/org_members.go:962-967`): `owned` is keyed on lowercase `Hex()` but looked up with the caller's spelling, so an uppercase-hex id is silently dropped. Not reachable through the API today (handlers pre-filter consistently), but the split is real inside `DeleteOrgMembers`.
- **Narrowing a whole-census question reports `removed: 0`** — the safety half is closed by the 409, but the reporting half is the exact response the PR's own verification section cites as the bug symptom; at minimum give `Added`/`Removed` godoc saying they diff the *named* list.
- **CSP swagger prose is behind the new refusals**: the `ProcessSignHandler`/`ProcessAuthHandler` 401 descriptions omit "question not open"/"not published"/"no longer part of the census", and `BundleSignHandler` can now answer a 404 (40029) its `@Failure 404` doesn't cover. The generated file matches the annotations; the annotations are stale.

Also verified and worth stating: the READY/PAUSED constant sets match between the guard and the CSP gate (CSP-signable ⊂ guard-ongoing, the safe direction); both sign handlers fail closed on DB errors; org scoping is complete on every `/processes`-model cascade entry point; `cspTokensStatus` genuinely survives every revocation path; and the stated on-chain ceiling is accurate against `cspproof.Verify` in dvote.

elboletaire (approved):
My agent messed up requesting changes here when it shouldn't

PR #622, opened 2026-07-29
feat(api): verify vote nullifiers at POST /votes/verify

lucasmenendez:
## Why

A voter relaying through `POST /vote` or `POST /votes` gets a nullifier back on the job, but had no way to confirm the ballot actually landed on chain. The SaaS API is the remote signer and otherwise hides the Vochain, so a voter-facing receipt / "verify my vote" screen had no endpoint to call.

## What

`POST /votes/verify` (public):

```jsonc
// request
{ "nullifiers": ["a1b2…", "c3d4…"] }

// 200
{ "votes": [
  { "nullifier": "a1b2…", "verified": true,
    "processId": "0f…", "txHash": "9e…", "blockHeight": 1423, "date": "2026-07-30T…" },
  { "nullifier": "c3d4…", "verified": false }
] }
```

One entry per requested nullifier, in request order. Several per call because a multi-question voting process is one on-chain election per question, so a voter holds one nullifier per question — the same reason `POST /votes` exists. The cap is the shared `maxVotesPerBatch`, so a voter can always verify in one call exactly what they were allowed to cast.

## Notes for review

- **Which node route.** `Account.VoteByNullifier` wraps the node's `GET /votes/{voteId}`, which is keyed on the nullifier alone — unlike its `/votes/verify/{electionId}/{voteId}`, which also needs the election id a caller holding only a nullifier does not have. It returns the receipt metadata as a bonus. No `apiclient` wrapper exists for it, hence the raw `Request`; a 404 becomes the `ErrVoteNotFound` sentinel.
- **Unknown vs. unreachable.** An unknown nullifier is `verified: false`. A chain that cannot be asked fails the whole call with 50004 — a voter must never be told their vote is missing when it merely could not be looked up.
- **No job, deliberately.** It answers synchronously and takes no `txQueue` slot. Those 16 workers block until a tx is *mined*; spending them on reads would let verification traffic starve real vote submissions. It also keeps a public unauthenticated endpoint from writing a job document per call.
- **Latency.** A vote lookup measures ~250ms against `api-dev.vocdoni.net`, so at the batch cap a sequential loop would be a ~25s request. The reads fan out through the existing `parallelForEach` (8 workers, `api/processes.go`) → ~3.2s worst case, ~250ms for the typical single nullifier. A cancelled request context stops the remaining reads.
- **Public group**, like `/vote`, `/votes` and `GET /jobs/{jobId}`: it exposes only data already public on chain.
- No new error codes and no new dependency — reuses `decodeCappedJSON`, `ErrVoteBatchEmpty`/`TooLarge`, `ErrMalformedBody`, `ErrVochainRequestFailed`.

## Test

`TestRelayVote` is extended: it already casts a real CSP-proofed vote on the containerized Vocone chain through `POST /vote`, and now verifies that nullifier (expects `verified`, matching `processId`, non-empty `txHash`), a random one (expects `verified: false`), and a 2-byte one (expects 400).

```
go test -race -run 'TestRelayVote$' ./api/   # pass
make lint                                    # clean
make swagger                                 # regenerated
```

### DIFF
--- a/account/process.go
+++ b/account/process.go
@@ -2,15 +2,18 @@ package account
 
 import (
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"math"
+	"net/http"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/vocdoni/saas-backend/db"
 	"github.com/vocdoni/saas-backend/internal"
 	"go.vocdoni.io/dvote/api"
+	"go.vocdoni.io/dvote/apiclient"
 	"go.vocdoni.io/proto/build/go/models"
 )
 
@@ -265,6 +268,35 @@ func (a *Account) Election(processID []byte) (*api.Election, error) {
 		return nil, fmt.Errorf("could not fetch election %x: %w", processID, err)
 	}
 	return election, nil
+}
+
+// ErrVoteNotFound is returned by VoteByNullifier when the chain has no vote with the
+// given nullifier. It is a distinct sentinel because "the chain does not know this vote"
+// is an answer, while any other failure means the chain could not be asked.
+var ErrVoteNotFound = fmt.Errorf("vote not found")
+
+// VoteByNullifier fetches an on-chain vote by its nullifier (voteID) from the Vochain,
+// returning ErrVoteNotFound when the chain does not know it. The node keys this lookup on
+// the nullifier alone — unlike its /votes/verify route, which also needs the election id —
+// so a caller holding only a nullifier can still resolve the vote and the election it
+// belongs to. There is no apiclient wrapper for it, hence the raw request.
+func (a *Account) VoteByNullifier(nullifier []byte) (*api.Vote, error) {
+	resp, code, err := a.client.Request(apiclient.HTTPGET, nil, "votes", hex.EncodeToString(nullifier))
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch vote %x: %w", nullifier, err)
+	}
+	switch code {
+	case http.StatusOK:
+		vote := &api.Vote{}
+		if err := json.Unmarshal(resp, vote); err != nil {
+			return nil, fmt.Errorf("could not decode vote %x: %w", nullifier, err)
+		}
+		return vote, nil
+	case http.StatusNotFound:
+		return nil, ErrVoteNotFound
+	default:
+		return nil, fmt.Errorf("could not fetch vote %x: status %d (%s)", nullifier, code, resp)
+	}
 }
 
 // ElectionEncryptionKeys fetches the encryption public keys of the on-chain election with
--- a/account/vote_test.go
+++ b/account/vote_test.go
@@ -1,0 +1,64 @@
+package account
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+	"go.vocdoni.io/dvote/api"
+	"go.vocdoni.io/dvote/apiclient"
+)
+
+// TestVoteByNullifier exercises the three answers a node can give — a vote, a 404, and any
+// other status — against a fake node, since the containerized chain cannot be made to fail
+// on demand. The failure case uses 500, not 503: apiclient retries 503 waiting for a block.
+func TestVoteByNullifier(t *testing.T) {
+	c := qt.New(t)
+
+	known := []byte(strings.Repeat("k", 32))
+	knownVote := &api.Vote{
+		ElectionID:  []byte(strings.Repeat("e", 32)),
+		TxHash:      []byte(strings.Repeat("t", 32)),
+		BlockHeight: 42,
+	}
+	broken := []byte(strings.Repeat("b", 32))
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case "/votes/" + hex.EncodeToString(known):
+			w.Header().Set("Content-Type", "application/json")
+			if err := json.NewEncoder(w).Encode(knownVote); err != nil {
+				t.Errorf("encoding vote: %v", err)
+			}
+		case "/votes/" + hex.EncodeToString(broken):
+			http.Error(w, "boom", http.StatusInternalServerError)
+		default: // including the /chain/info probe apiclient.New makes
+			http.Error(w, "not found", http.StatusNotFound)
+		}
+	}))
+	defer srv.Close()
+
+	// the trailing slash matters: apiclient joins paths onto the base URL's path, and an
+	// empty base path would yield a request line without a leading slash (a 400 from net/http)
+	client, err := apiclient.New(srv.URL + "/")
+	c.Assert(err, qt.IsNil)
+	a := &Account{client: client}
+
+	vote, err := a.VoteByNullifier(known)
+	c.Assert(err, qt.IsNil)
+	c.Assert([]byte(vote.ElectionID), qt.DeepEquals, []byte(knownVote.ElectionID))
+	c.Assert([]byte(vote.TxHash), qt.DeepEquals, []byte(knownVote.TxHash))
+	c.Assert(vote.BlockHeight, qt.Equals, knownVote.BlockHeight)
+
+	_, err = a.VoteByNullifier([]byte(strings.Repeat("u", 32)))
+	c.Assert(err, qt.ErrorIs, ErrVoteNotFound)
+
+	_, err = a.VoteByNullifier(broken)
+	c.Assert(err, qt.Not(qt.ErrorIs), ErrVoteNotFound, qt.Commentf("a node failure must not read as not-found"))
+	c.Assert(err, qt.ErrorMatches, fmt.Sprintf("(?s)could not fetch vote %x: status 500.*", broken))
+}
--- a/api/api.go
+++ b/api/api.go
@@ -462,6 +462,7 @@ func (a *API) initRouter() http.Handler {
 		handle(r, http.MethodGet, processEndpoint, a.processInfoHandler)
 		handle(r, http.MethodPost, voteEndpoint, a.relayVoteHandler)
 		handle(r, http.MethodPost, votesEndpoint, a.relayVotesHandler)
+		handle(r, http.MethodPost, votesVerifyEndpoint, a.verifyVotesHandler)
 		handle(r, http.MethodGet, processResultsEndpoint, a.processResultsHandler)
 		handle(r, http.MethodGet, processMetadataEndpoint, a.processMetadataHandler)
 		handle(r, http.MethodPost, processSignInfoEndpoint, cspHandlers.ConsumedAddressHandler)
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -1327,7 +1327,8 @@ type RelayVotesRequest struct {
 // voting process.
 // swagger:model VerifyVotesRequest
 type VerifyVotesRequest struct {
-	// Vote nullifiers (32 bytes each), at most 100
+	// Vote nullifiers (up to 32 bytes each — anonymous ones may be shorter), capped at the
+	// shared vote batch limit (100)
 	Nullifiers []internal.HexBytes `json:"nullifiers" swaggertype:"array,string"`
 }
 
--- a/api/apicommon/types.go
+++ b/api/apicommon/types.go
@@ -1322,6 +1322,40 @@ type RelayVotesRequest struct {
 	Votes []RelayVoteRequest `json:"votes"`
 }
 
+// VerifyVotesRequest is the body of POST /votes/verify: the vote nullifiers a voter wants
+// checked against the chain, typically the one nullifier per question of a multi-question
+// voting process.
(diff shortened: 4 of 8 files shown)


elboletaire (changes requested):

Reviewed at `f7ac888`. Verified locally: `go build ./...` + `go vet` clean, `golangci-lint run ./api/... ./account/...` reports no issues, and `go test -race -run 'TestRelayVote$' ./api/` passes (23.9s).

The design holds up against the code. The node's `getVoteHandler` really does return `ErrVoteNotFound` with HTTP 404 (`dvote/api/errors.go:52`), so the 404 → sentinel mapping is sound and every other status genuinely means "could not ask". `/votes/{voteId}` is keyed on the nullifier alone, so the choice of node route is right. The handler discards the node's `VotePackage`, `VoterID`, `weight` and `overwriteCount`, so no ballot content or voter identity leaks. The body is capped before buffering, both batch guards run before any chain read, and the goroutines write disjoint indices — `-race` agrees no locking is needed. Findings below, ordered by severity.

---

## Blocking

### 1. The strict 32-byte nullifier check rejects valid anonymous nullifiers

`api/process_vote.go:288-293`

For anonymous (ZK) votes the nullifier is not a fixed-width hash. `zk.Proof.Nullifier()` returns `bi.Bytes()` (`dvote/crypto/zk/prover/pubsignals.go:98`) — the minimal big-endian encoding of a field element, so ≤ 32 bytes, and roughly 1 in 256 values is 31 bytes or shorter. That path is reachable through this backend:

- `account/process.go:157` maps `ep.ElectionType.Anonymous` into the on-chain `EnvelopeType`, and `subscriptions/subscriptions.go:93` gates it as a plan feature — so anonymous elections do get created here;
- `dvote/vochain/transaction/vote_tx.go:121-130` routes `OFF_CHAIN_CA` (the census origin this backend always uses) **plus** `Anonymous` to `zkproof.InitializeZkVote`;
- `voteNullifier` (`api/process_vote.go:426-428`) passes `vote.Nullifier` through verbatim, and that same value is what `POST /vote` returns on the job.

So the API can hand a voter a 31-byte nullifier and then reject it with 400 on `/votes/verify`. Padding is not the fix — the indexer stores the trimmed bytes, so the value has to be passed through as-is. Match the node, which only rejects empty:

```go
if len(nullifier) == 0 || len(nullifier) > nullifierSize {
	errors.ErrMalformedBody.Withf("the nullifier at index %d is %d bytes, expected 1..%d",
		i, len(nullifier), nullifierSize).Write(w)
	return
}
```

The `nullifierSize` comment needs adjusting too: `state.GenerateNullifier` is a 32-byte hash, but that is only the non-anonymous derivation.

---

## Should fix

### 2. Node error text reaches unauthenticated callers, up to 100 of them

`errors.Error.MarshalJSON` (`errors/errors.go:26-38`) serializes `Err.Error()` into the response body, so `ErrVochainRequestFailed.WithErr(stderrors.Join(errs...))` returns the raw node response bodies (`could not fetch vote %x: status %d (%s)`) concatenated, on a public endpoint. Using `WithErr` for chain failures matches the existing handlers, so the pattern is fine — joining up to 100 of them is what is new. Suggest logging the joined detail and returning the bare sentinel, or wrapping only the first non-nil error.

Related: when `middleware.Timeout(45 * time.Second)` fires, the join becomes a pile of `context deadline exceeded` — a fine outcome, but not a useful body.

### 3. Public 1→100 read amplification against a shared node

The dedup closes the repeat-one-value trick, but 100 *distinct* nullifiers still turn one unauthenticated request into 100 node GETs. `router()` only has `middleware.Throttle(100)` / `ThrottleBacklog` — global concurrency, not a per-IP rate limit — so the worst case is roughly 800 concurrent and 10k queued node reads. The relay endpoints are public too, but their writes are bounded by the shared 16-worker `txQueue`; here each request brings its own pool of 8, so the ceiling scales with request concurrency. Worth one of:

- a package-level semaphore shared across requests, rather than per-request;
- a short positive cache — a mined vote's answer is immutable, and a voter-facing "verify my vote" screen is exactly the polling workload that benefits (`resolveQuestionEncryptionKeys` already establishes the cache-on-chain-read pattern);
- a lower cap for this endpoint than the relay batch cap.

The amplifier also interacts badly with `apiclient`'s retry loop: a node replying 503 makes `Request` call `WaitUntilNextBlock()` up to `DefaultRetries = 3` times (`dvote/apiclient/client.go:204-223`), so a single lookup can block well past the 45s timeout. Pre-existing for every chain read in the codebase, but the 100× fan-out is what makes it visible.

---

## Consider (design decisions, fine to ship either way)

### 4. All-or-nothing on partial chain failure

One flaky lookup out of ten fails the whole call, so a voter with ten questions is told the chain is unavailable even though nine verified. The docstring's own reasoning — never tell a voter their vote is missing when it merely could not be looked up — argues for a per-entry third state rather than a batch-level 500, e.g. `status: verified|unverified|unknown` instead of a bool. Cheap now, breaking later, so worth deciding consciously.

### 5. `processId` naming

The field is the on-chain election id, while `{processId}` everywhere else in this API is the 24-hex SaaS ProcessID (`api/routes.go:169,185,189`). A client cannot feed the returned value to `GET /process/{processId}/results`. It matches `db.JobResult.processId` from the relay path, so it is internally consistent — but `electionId` (the node's own name) would be less ambiguous, and for multi-question processes the value is the *question's* upstream id, not the process's.

### 6. Scope: any nullifier is accepted

Including votes in elections this backend does not manage, which makes the SaaS a free read proxy for the node. Everything is public on chain, so this is not a disclosure issue. A `QuestionByUpstreamID` / `ProcessByAddress` filter on `vote.ElectionID` would narrow it and make `processId` resolvable, at the cost of one indexed DB read per hit.

---

## Nits

- **Swagger drift:** `VerifyVotesRequest.Nullifiers` hardcodes "at most 100" while the cap is `maxVotesPerBatch = maxQuestionsPerProcess`.
- `BlockHeight uint32` with `omitempty` would drop a legitimate block 0 — impossible in practice, ignore.
- `firstOf` + `distinct` reads a little heavy for what it does, but the comment earns it; no change needed.
- `var ErrVoteNotFound` inline at `account/process.go:276` is the package's only sentinel; placement is fine.

---

## Test coverage

The `TestRelayVote` extension is well targeted: a real chain-backed positive, an unknown nullifier, a duplicate, a malformed length, an empty batch and an over-cap batch — and the over-cap body at ~6.7 KB stays under the 12 KB reader cap, so it correctly lands on 40165 rather than 413. Gaps:

- no coverage of the node-error → 50004 path, which is the whole point of the unknown-vs-unreachable distinction;
- no unit test for `Account.VoteByNullifier`'s 404 and default branches;
- duplicates are asserted to return equal answers, but nothing proves only one chain read happened (needs a fake to observe);
- no anonymous-election case, which is where finding 1 bites.

## Conventions

Compliant throughout: `fmt.Errorf` with `%w`, lowercase message starts, godoc on every exported entity starting with the entity name, swagger annotations regenerated into `docs/swagger.yaml`, and `qt` matchers (`qt.HasLen`, `qt.IsTrue`, `qt.DeepEquals`, `qt.Not(qt.HasLen)`) rather than `len()` + `Equals`. Comment density matches the surrounding file.

---

Requesting changes on finding 1 only — it makes the endpoint reject nullifiers the same API issued, for anonymous elections. Finding 2 is a two-line change, and findings 3 and 4 mostly need a recorded decision. Everything else is take-or-leave.

lucasmenendez:
Thanks for the thorough review — the anonymous-nullifier catch especially. Addressed in 9a376b0:

**1. Short anonymous nullifiers (blocking) — fixed.** Verified the claim against `zk.Proof.Nullifier()` (`bi.Bytes()`, minimal big-endian): a ≤31-byte nullifier is real and this API could hand one out and then 400 it. The check is now `len == 0 || len > 32`, matching the node, with the error message and the `nullifierSize` comment rewritten to say 32 is the maximum (hash derivation) and shorter ZK values are valid. The test now asserts a 2-byte nullifier is accepted (and reported unverified), and the malformed cases are empty and 33 bytes.

**2. Node error text on a public endpoint — fixed.** The joined per-nullifier detail now goes to the log (`log.Warnw`) and the caller gets the bare 50004 sentinel.

**3. Read amplification — fixed with a shared semaphore.** A package-level 32-slot semaphore now bounds concurrent node reads across *all* verify requests, so the ceiling no longer scales with request concurrency (was ~800 under `Throttle(100)`; now 32, i.e. four fully-fanned batches). Workers wait on the slot or the request context, whichever comes first. I went with the semaphore over the positive cache because it's the direct fix for the scaling complaint with no invalidation questions; the cache remains a good follow-up if verify polling becomes a measurable load.

**4. All-or-nothing on partial failure — keeping it, deliberately.** The "cheap now, breaking later" concern is softer than it looks: a per-entry `status` field can be *added* later alongside `verified` without breaking anyone. Until then, fail-the-call keeps the voter-screen contract trivial (retry the whole thing), and `apiclient`'s own retries already absorb most transient flake.

**5. `processId` naming — keeping it.** It's the same value, under the same name, that the relay job result gives the voter (`JobResult.processId`) — that's the pairing a client actually correlates. Renaming here would make the two halves of the same flow disagree.

**6. Open scope — keeping it.** A DB filter on `vote.ElectionID` would run *after* the chain read, so it saves the node nothing; it would only narrow what we echo back, at one indexed read per hit, for data that's public on chain anyway.

**Nits:** the swagger comment no longer hardcodes a bare "32 bytes each"/"at most 100" — it now says up to 32 bytes (anonymous may be shorter) and names the cap as the shared vote batch limit.

**Test gaps:** added `TestVoteByNullifier` (`account/vote_test.go`) covering the 200/404/500 branches against a fake node — including that a node failure never reads as not-found, which is the unknown-vs-unreachable distinction at the unit level (and should recover the `account/` coverage dip). The handler-level 50004 path and a proof that duplicates cost one read would both need a fake account injected into the API under test, which the harness doesn't support today — noted as a gap rather than forced in. A real anonymous-election E2E needs ZK proof generation against Voconed; the unit-level acceptance of short nullifiers is covered in `TestRelayVote`.

elboletaire (approved):

Re-reviewed at `9a376b0` (the rebase is clean — `git range-diff` shows the two original commits byte-identical to their pre-rebase versions, so the fix commit is the only new work).

**Verified locally:** `go build ./...` + `go vet` clean · `golangci-lint run ./api/... ./account/...` no issues · `make swagger` produces no diff, so the generated docs are in sync · `go test -race -run TestVoteByNullifier ./account/` passes · `go test -race -run 'TestRelayVote$|TestRelayVotesBatch$' ./api/` passes (41.4s).

## Blocking finding — fixed

**1. Short anonymous nullifiers.** Now `len(nullifier) == 0 || len(nullifier) > nullifierSize`, matching the node. The `nullifierSize` comment correctly reframes 32 as the maximum and names both derivations (`state.GenerateNullifier` hash vs. `zk.Proof.Nullifier` field element), and the error message reads `expected 1..32`. The test change is the right shape: a 2-byte nullifier is asserted *accepted* and reported unverified, with empty and 33 bytes as the malformed cases. I checked the empty case round-trips as intended rather than short-circuiting — `HexBytes.MarshalJSON` emits `""`, which clears `UnmarshalJSON`'s `len(data) < 2` guard and decodes to a zero-length value, so the `len == 0` branch is genuinely exercised.

## Should-fix findings — both fixed

**2. Node error text on a public endpoint.** Joined detail to `log.Warnw`, caller gets the bare 50004. Correct.

**3. Read amplification.** I checked the shared semaphore for the failure mode it could have introduced, and it's sound: `parallelForEach` takes its local 8-slot semaphore in the caller goroutine, workers then take the global one, and a goroutine holding a global slot never waits on a local slot — no cycle, so no deadlock. The `select` on the slot vs. `r.Context().Done()` means parked workers bail when the request dies instead of pinning slots, and the releasing `defer` is function-scoped so it fires on callback return. `-race` clean across both integration tests.

One residual, not a reopen: the semaphore bounds *concurrency*, not *rate*. Under a flood the choke point moves from the node to this service — 32 slots at ~250ms is ~128 reads/s, so 100-nullifier requests take ~78s, exceed the 45s `middleware.Timeout`, and hold `Throttle(100)` slots while doing it, crowding out login/relay traffic. Still the right trade, since the node is shared across orgs while this API is the layer you can scale and rate-limit per deployment — the real fix is per-IP rate limiting, which is out of scope here. Worth a follow-up issue rather than a change to this PR.

## Declines — reasoning holds

- **4 (all-or-nothing):** right that a per-entry `status` alongside `verified` can be added later without breaking anyone, so the "breaking later" concern doesn't bind. Your call, and the retry-the-whole-thing contract is defensible.
- **5 (`processId` naming):** fair — this field and `JobResult.processId` are the two halves of one flow and a client correlates them; renaming one would make them disagree.
- **6 (open scope):** correct on the facts, the DB filter would run *after* the chain read and save the node nothing. My point was scope-narrowing rather than load, and declining it is reasonable for data that is public on chain anyway.

## Test coverage

`TestVoteByNullifier` covers all three node answers, and asserting that a 500 never reads as `ErrVoteNotFound` is exactly the right unit-level shape for the unknown-vs-unreachable distinction. The stated reason for the two remaining gaps checks out: `api.API.account` is a concrete `*account.Account` (`api/api.go:157`), so there is no seam to inject a failing or read-counting fake without adding an interface — correctly left as a noted gap rather than forced in.

Approving. Nice work on the turnaround, and thanks for verifying the ZK nullifier claim independently rather than taking it on faith.

PR #626, opened 2026-07-30
fix(api)!: remove the organization `active` flag

### LINKED ISSUE #625: Organization `active` flag: behavior is undefined — investigate, then remove or properly define it
This is Claude (Fable 5) speaking on behalf of elboletaire.

## Description

While reviewing the vocdoni-app migration to the integrator-sdk, we tried to use the `active` flag returned on the `Organization` object (`GET /organizations/{address}`, `apicommon.OrganizationInfo`) to gate the dashboard for suspended/archived organizations. It immediately broke for a real account: an organization that is fully usable — its owner can edit every organization setting **and create new voting processes without any issue** — is returned with `active: false`, while other organizations of the same user have `active: true`.

The name strongly suggests that an org with `active: false` should not be able to change anything or operate at all, yet nothing in the backend appears to enforce it: writes, process creation and publishing all succeed regardless of the flag's value.

## Current behavior

- `active` is exposed on the organization read endpoints and typed in `@vocdoni/api-types`.
- Its value does not correlate with whether the organization is operational: at least one org with `active: false` can be edited freely and can create/publish processes.
- No known endpoint rejects operations based on it, and there is no documented way for it to transition between values.

## Expected

Either of:

1. **Define it properly** — document what `active` means, what sets/unsets it, and enforce it consistently on the write/process-creation paths so clients can rely on it (e.g. to hide the dashboard for suspended orgs), or
2. **Remove it** — if it is a leftover with no enforced semantics, drop it from the API responses and from `@vocdoni/api-types` so clients don't build gating on top of a meaningless field.

## Context

- Found while addressing a vocdoni-app peer-review finding that wanted archived/suspended-org gating wired to this flag; the gating had to be reverted because the flag misreported a healthy org (branch `refactor/integrator-sdk-remove-remote-signer`).
- Worth checking where `active` is set in the org lifecycle (creation, subscription changes, deletion/archival) to understand why sibling orgs of the same user end up with different values.

lucasmenendez:
Closes #625.

## What was wrong

`Organization.active` was exposed on the organization read endpoints, but **nothing read it**. The only non-test consumer of `org.Active` was the JSON mapping in `OrganizationFromDB` — no middleware, no handler guard, no subscription check. An organization stored with `active: false` remained fully editable and could create and publish processes, exactly as reported.

Nor did anything set it deliberately:

- both creation paths hardcoded `Active: true` and silently discarded the value sent in the request body;
- there is no archival — `DelOrganization` hard-deletes;
- no migration ever touched it.

**The `false` values came from a bug.** `PUT /organizations/{address}` is a partial update: every other field is guarded by an emptiness check (`if newOrgInfo.Website != ""`). `Active` was a plain `bool`, so a request body omitting `active` decoded to `false`, differed from the stored `true`, and flipped the organization. That write was then force-persisted, because `"active"` was the single field opted out of zero-value protection in `SetOrganization`:

```go
// define 'active' parameter to be updated always to update it even its new value is false
updateDoc, err := dynamicUpdateDocument(org, []string{"active"})
```

So any client updating one unrelated setting without echoing `active` back deactivated the org — which is why sibling orgs of the same user disagree. `TestOrganizationPartialUpdate` sends that exact request shape but never asserted `Active` survived, so nothing caught it.

## What this does

Option 2 of the issue: **remove it.** Defining it would mean inventing a suspension lifecycle, an authority model (today any org admin could suspend themselves), and a guard on every write handler — none of which exists or was asked for. Removing the field eliminates the bug by construction.

- Dropped from `db.Organization`, `apicommon.OrganizationInfo` (and therefore from both create-request wrappers), `OrganizationFromDB`, both creation handlers, and the `PUT` handler.
- `SetOrganization` no longer force-writes any zero value.
- **Migration `0019_drop_organization_active`** `$unset`s the stale field; the rollback restores `active: true`, the only value the code ever wrote deliberately.
- Regression test asserts the organization payload carries no `active` key; migration test covers up and down.
- Swagger regenerated.

`Subscription.Active` (Stripe-driven, gates `subscriptions.IsIntegrator`) is a **different field and is untouched**.

## ⚠️ Breaking change

`active` disappears from organization responses and create requests. It must be dropped from `@vocdoni/api-types`. No known consumer reads it — the integrator demo has zero references, and the vocdoni-app gating that prompted this issue was already reverted.

## Verification

- `go build ./... && go vet ./...` clean
- `golangci-lint run ./db/... ./api/... ./migrations/...` → 0 issues
- `go test ./api/... ./db/ ./csp/... ./subscriptions/` → all pass, including the new `TestDropOrganizationActiveMigration`

### DIFF
--- a/api/organizations.go
+++ b/api/organizations.go
@@ -143,7 +143,6 @@ func (a *API) createOrganizationHandler(w http.ResponseWriter, r *http.Request) 
 		Country:         orgInfo.Country,
 		Subdomain:       orgInfo.Subdomain,
 		Timezone:        orgInfo.Timezone,
-		Active:          true,
 		Communications:  orgInfo.Communications,
 		Meta:            apicommon.BuildOrgMeta(nil, orgInfo.Name, orgInfo.Logo, orgInfo.Description, orgInfo.Meta),
 		TokensPurchased: 0,
@@ -280,10 +279,6 @@ func (a *API) updateOrganizationHandler(w http.ResponseWriter, r *http.Request) 
 	}
 	if newOrgInfo.Timezone != "" {
 		org.Timezone = newOrgInfo.Timezone
-		updateOrg = true
-	}
-	if newOrgInfo.Active != org.Active {
-		org.Active = newOrgInfo.Active
 		updateOrg = true
 	}
 	if newOrgInfo.Name != nil || newOrgInfo.Logo != nil || newOrgInfo.Description != nil || len(newOrgInfo.Meta) > 0 {
--- a/api/organizations_test.go
+++ b/api/organizations_test.go
@@ -166,19 +166,15 @@ func TestUpdateOrganizationHandler(t *testing.T) {
 	_, code = testRequest(t, http.MethodPut, token, updateInfo, "organizations", nonExistentAddr.String())
 	c.Assert(code, qt.Equals, http.StatusBadRequest)
 
-	// Test updating active status
-	activeUpdateInfo := &apicommon.OrganizationInfo{
-		Active: false,
-	}
-	resp, code = testRequest(t, http.MethodPut, token, activeUpdateInfo, "organizations", orgAddress.String())
-	c.Assert(code, qt.Equals, http.StatusOK, qt.Commentf("response: %s", resp))
-
-	// Verify active status update
+	// The organization payload must not expose an `active` flag (issue #625): nothing enforced
+	// it, and because it was a plain bool on a partial update, any request omitting it silently
+	// deactivated the organization.
 	resp, code = testRequest(t, http.MethodGet, token, nil, "organizations", orgAddress.String())
 	c.Assert(code, qt.Equals, http.StatusOK)
-	var orgWithUpdatedStatus apicommon.OrganizationInfo
-	c.Assert(json.Unmarshal(resp, &orgWithUpdatedStatus), qt.IsNil)
-	c.Assert(orgWithUpdatedStatus.Active, qt.IsFalse)
+	rawOrg := map[string]any{}
+	c.Assert(json.Unmarshal(resp, &rawOrg), qt.IsNil)
+	_, hasActive := rawOrg["active"]
+	c.Assert(hasActive, qt.IsFalse, qt.Commentf("response: %s", resp))
 }
 
 func TestOrganizationsTypesHandler(t *testing.T) {
@@ -487,7 +483,6 @@ func TestOrganizationWithOptionalFields(t *testing.T) {
 		Country:        "ES",
 		Subdomain:      "fullexample",
 		Timezone:       "Europe/Madrid",
-		Active:         true,
 		Communications: true,
 	}
 	resp, code := testRequest(t, http.MethodPost, token, orgInfo, organizationsEndpoint)
@@ -503,7 +498,6 @@ func TestOrganizationWithOptionalFields(t *testing.T) {
 	c.Assert(createdOrg.Country, qt.Equals, "ES")
 	c.Assert(createdOrg.Subdomain, qt.Equals, "fullexample")
 	c.Assert(createdOrg.Timezone, qt.Equals, "Europe/Madrid")
-	c.Assert(createdOrg.Active, qt.IsTrue)
 	c.Assert(createdOrg.Communications, qt.IsTrue)
 }
 
--- a/api/organizations_test.go
+++ b/api/organizations_test.go
@@ -148,6 +148,12 @@ func TestUpdateOrganizationHandler(t *testing.T) {
 	c.Assert(updatedOrg.Color, qt.Equals, "#FF5733")
 	c.Assert(updatedOrg.Size, qt.Equals, "medium")
 
+	// The payload must not expose an `active` flag (issue #625).
+	rawOrg := map[string]any{}
+	c.Assert(json.Unmarshal(resp, &rawOrg), qt.IsNil)
+	_, hasActive := rawOrg["active"]
+	c.Assert(hasActive, qt.IsFalse, qt.Commentf("response: %s", resp))
+
 	// Test updating without authentication
 	_, code = testRequest(t, http.MethodPut, "", updateInfo, "organizations", orgAddress.String())
 	c.Assert(code, qt.Equals, http.StatusUnauthorized)
@@ -165,16 +171,6 @@ func TestUpdateOrganizationHandler(t *testing.T) {
 	nonExistentAddr := common.HexToAddress("0x0000000000000000000000000000000000000001")
 	_, code = testRequest(t, http.MethodPut, token, updateInfo, "organizations", nonExistentAddr.String())
 	c.Assert(code, qt.Equals, http.StatusBadRequest)
-
-	// The organization payload must not expose an `active` flag (issue #625): nothing enforced
-	// it, and because it was a plain bool on a partial update, any request omitting it silently
-	// deactivated the organization.
-	resp, code = testRequest(t, http.MethodGet, token, nil, "organizations", orgAddress.String())
-	c.Assert(code, qt.Equals, http.StatusOK)
-	rawOrg := map[string]any{}
-	c.Assert(json.Unmarshal(resp, &rawOrg), qt.IsNil)
-	_, hasActive := rawOrg["active"]
-	c.Assert(hasActive, qt.IsFalse, qt.Commentf("response: %s", resp))
 }
 
 func TestOrganizationsTypesHandler(t *testing.T) {
--- a/db/migration_drop_org_active_test.go
+++ b/db/migration_drop_org_active_test.go
@@ -9,7 +9,7 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 )
 
-// TestDropOrganizationActiveMigration asserts migration 0019 strips the organizations' `active`
+// TestDropOrganizationActiveMigration asserts migration 0020 strips the organizations' `active`
 // field, and that the rollback restores it as true (issue #625).
 func TestDropOrganizationActiveMigration(t *testing.T) {
 	c := qt.New(t)
@@ -25,7 +25,7 @@ func TestDropOrganizationActiveMigration(t *testing.T) {
 		bson.M{"_id": testOrgAddress}, bson.M{"$set": bson.M{"active": false}})
 	c.Assert(err, qt.IsNil)
 
-	mig := migrationByVersion(c, 19)
+	mig := migrationByVersion(c, 20)
 	c.Assert(mig.Up(ctx, database), qt.IsNil)
 
 	raw := bson.M{}
--- a/db/migration_drop_org_active_test.go
+++ b/db/migration_drop_org_active_test.go
@@ -6,12 +6,11 @@ import (
 	"time"
 
 	qt "github.com/frankban/quicktest"
-	"github.com/vocdoni/saas-backend/migrations"
 	"go.mongodb.org/mongo-driver/bson"
 )
 
-// TestDropOrganizationActiveMigration checks that migration 0019 removes the `active` field left
-// behind by pre-#625 writers, including the false values a partial update wrote by accident.
+// TestDropOrganizationActiveMigration asserts migration 0019 strips the organizations' `active`
+// field, and that the rollback restores it as true (issue #625).
 func TestDropOrganizationActiveMigration(t *testing.T) {
 	c := qt.New(t)
 	c.Cleanup(func() { c.Assert(testDB.DeleteAllDocuments(), qt.IsNil) })
@@ -26,8 +25,7 @@ func TestDropOrganizationActiveMigration(t *testing.T) {
 		bson.M{"_id": testOrgAddress}, bson.M{"$set": bson.M{"active": false}})
 	c.Assert(err, qt.IsNil)
 
-	mig, ok := migrations.AsMap()[19]
-	c.Assert(ok, qt.IsTrue)
+	mig := migrationByVersion(c, 19)
 	c.Assert(mig.Up(ctx, database), qt.IsNil)
 
 	raw := bson.M{}
@@ -36,6 +34,16 @@ func TestDropOrganizationActiveMigration(t *testing.T) {
 	c.Assert(hasActive, qt.IsFalse)
 
 	// The rollback restores the only value the code ever wrote deliberately.
+	c.Assert(mig.Down(ctx, database), qt.IsNil)
+	raw = bson.M{}
+	c.Assert(testDB.organizations.FindOne(ctx, bson.M{"_id": testOrgAddress}).Decode(&raw), qt.IsNil)
+	c.Assert(raw["active"], qt.IsTrue)
+
+	// A false surviving a partial up, or written by an old binary mid-deploy, is overwritten too:
+	// the rollback restores true unconditionally, not only where the field is missing.
+	_, err = testDB.organizations.UpdateOne(ctx,
+		bson.M{"_id": testOrgAddress}, bson.M{"$set": bson.M{"active": false}})
+	c.Assert(err, qt.IsNil)
 	c.Assert(mig.Down(ctx, database), qt.IsNil)
 	raw = bson.M{}
 	c.Assert(testDB.organizations.FindOne(ctx, bson.M{"_id": testOrgAddress}).Decode(&raw), qt.IsNil)
--- a/db/migration_drop_org_active_test.go
+++ b/db/migration_drop_org_active_test.go
@@ -1,0 +1,43 @@
+package db
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	qt "github.com/frankban/quicktest"
+	"github.com/vocdoni/saas-backend/migrations"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+// TestDropOrganizationActiveMigration checks that migration 0019 removes the `active` field left
+// behind by pre-#625 writers, including the false values a partial update wrote by accident.
+func TestDropOrganizationActiveMigration(t *testing.T) {
+	c := qt.New(t)
+	c.Cleanup(func() { c.Assert(testDB.DeleteAllDocuments(), qt.IsNil) })
+
+	ctx := context.Background()
+	database := testDB.DBClient.Database(testDB.database)
+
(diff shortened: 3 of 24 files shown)


elboletaire (changes requested):

Verified locally on `95487a8`: `go build ./...`, `go vet ./...` and `scripts/check-qt-patterns.sh` clean; `TestDropOrganizationActiveMigration`, `TestMigrations`, `TestOrganization`, `TestUpdateOrganizationHandler`, `TestOrganizationWithOptionalFields` and `TestOrganizationPartialUpdate` all pass.

The change itself is right, and the diagnosis in the description matches the tree — `active` was the only entry in `dynamicUpdateDocument`'s `alwaysUpdateTags` for organizations, and the plain-`bool` comparison in the `PUT` handler is exactly how `false` got persisted. Requesting changes for one thing only: the docs are left half-updated.

## Requested change: finish the docs

`docs/swagger.yaml` was regenerated, but `api/docs.md` still documents the field this PR removes. That file is hand-maintained and kept current by recent PRs (`d383c87`, `f584b44`), so it is part of this repo's API surface — not generated output that can be regenerated later. Shipping a breaking removal while the docs still tell clients to send and expect `active` leaves dead documentation behind, and the *Update organization* example is the worst case: it documents sending the exact key that caused #625.

Occurrences to remove:

- `api/docs.md:564` — prose: "By default, the organization is created with `activated: true`."
- `api/docs.md:587` — *Create organization* response example
- `api/docs.md:643` — *Update organization* request body example
- `api/docs.md:674` — *Organization info* response example
- `api/docs.md:410` — *Get current user info*, embedded organization object

Leave `:420`, `:597`, `:684` and `:949` untouched — those are `subscription.active`, which is a different field.

## Everything else checks out

- **Nothing else left behind.** A repo-wide sweep for `active` turns up only `Subscription`/`OrganizationSubscription.Active` (`db/types.go:220`, `api/apicommon/types.go:881`, `stripe/*`) and prose about active *elections*/plans. No handler, index or Mongo `$jsonSchema` validator referenced the org field — `migrations/0001_initial_collections.go` defines no organizations validator, so the removal cannot trip document validation.
- **`dynamicUpdateDocument(org, nil)` is safe.** The only remaining `bool` on `Organization` is `Communications`, which was never in the force-update list, so its zero-value-skip behaviour is unchanged. The call now matches all twelve other callers in `db/`.
- **Migration is solid.** Both directions filter on `$exists` (idempotent, as `migrations/README.md` asks), errors are wrapped with `%w`, the log message is lowercase and reports `ModifiedCount`, and restoring `true` on rollback is the correct choice given both creation paths hardcoded it.
- **Request-side compatibility is better than the description suggests.** Go's JSON decoder ignores unknown fields, so old clients still sending `"active": true` in a `POST`/`PUT` body get a 200, not a 400. Only response *readers* break.
- **Test shape is right.** Asserting on a decoded `map[string]any` rather than the typed struct is what actually catches the key reappearing — a typed assertion would silently pass. The migration test covers both directions and seeds `active: false` directly, which is now the only way to reproduce pre-migration state.

## Two notes, no action needed

- **Rolling deploy.** While old instances are still serving, any `SetOrganization` they perform re-adds `active` after migration 19 has run, since it was in their force-update list. The values are inert — the driver ignores unknown BSON fields when decoding into the new struct — so the field merely reappears cosmetically until the deploy finishes. Worth a line in the release notes rather than a re-run.
- **Adjacent gap, out of scope.** `updateOrganizationHandler` (`api/organizations.go:259-286`) has no branch for `Communications` at all, so that field cannot be changed via `PUT` in either direction — and being a plain `bool` skipped as a zero value, it could not be set back to `false` even if a branch existed. Same class of problem as #625; worth its own issue.

No security or performance concerns: removing an unenforced flag weakens no authorization path (there was no guard to remove, which is the point of the issue), and the migration is one `UpdateMany` over a small collection at startup.

lucasmenendez:
Docs finished in 2a166a8 — removed the five organization-level `active` occurrences (`:410` user info, `:564` prose, `:587` create response, `:643` update request, `:674` org info). The four `subscription.active` examples are left untouched.

migrations/0019_drop_organization_active.go
```diff
@@ -0,0 +1,44 @@
...
+	"go.vocdoni.io/dvote/log"
+)
+
+func init() {
+	AddMigration(19, "drop_organization_active", upDropOrganizationActive, downDropOrganizationActive)
+}
+
+// upDropOrganizationActive removes the organizations' `active` field (issue #625). Nothing ever
+// read it: no handler, middleware or subscription check gated on it, so an organization stored
+// with active=false stayed fully operational. The only writer that ever set it to false was a bug
```
elboletaire: **[confirmed] The premise "The only writer that ever set it to false was a bug" is inaccurate, which makes the up migration lossier than documented.** The deleted handler branch (`if newOrgInfo.Active != org.Active`) honored an explicit, deliberate `PUT {"active": false}`, and the deleted test in `api/organizations_test.go` exercised exactly that round-trip — so explicit deactivation was supported, tested API behavior, not merely an artifact of the omitted-field bug. The `$unset` conflates deliberate and accidental `false` values with no record, and a later rollback resurrects any deliberate deactivation as `active: true`. Since nothing reads the flag the behavioral impact is nil, but if you can't confirm no production org was deliberately deactivated, consider having the up migration log (or archive) the addresses it strips `false` from, and soften the godoc claim.

migrations/0019_drop_organization_active.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	log.Infow("dropped organization active field", "organizations", res.ModifiedCount)
+	return nil
+}
+
+// downDropOrganizationActive restores the field as true, the only value the code ever wrote
+// deliberately: both creation paths hardcoded it, and every false was the bug described above.
+func downDropOrganizationActive(ctx context.Context, database *mongo.Database) error {
+	if _, err := database.Collection("organizations").UpdateMany(ctx,
+		bson.M{"active": bson.M{"$exists": false}},
```
elboletaire: **[plausible] The `$exists: false` filter contradicts the down's stated policy after a partial up.** If the up is interrupted (or a stray `active: false` is written between up and down by an old binary), the down leaves that `false` untouched while the comment claims it "restores the field as true" — reintroducing exactly the misreporting this PR removes. An unconditional `UpdateMany(ctx, bson.M{}, {$set: {active: true}})` matches the stated rollback policy and is no more expensive.

api/organizations.go
```diff
@@ -282,10 +281,6 @@ func (a *API) updateOrganizationHandler(w http.ResponseWriter, r *http.Request)
 		org.Timezone = newOrgInfo.Timezone
 		updateOrg = true
 	}
-	if newOrgInfo.Active != org.Active {
-		org.Active = newOrgInfo.Active
-		updateOrg = true
-	}
 	if newOrgInfo.Name != nil || newOrgInfo.Logo != nil || newOrgInfo.Description != nil || len(newOrgInfo.Meta) > 0 {
```
elboletaire, line 284: **[confirmed, pre-existing but in the touched block] `communications` is advertised as updatable but silently ignored.** The PUT body type (`apicommon.OrganizationInfo`) carries `communications`, and `api/docs.md` line 639 still lists `"communication": false` under “Only the following parameters can be changed” (this PR edited that exact example block to drop `"active": true` and left the stale claim). The handler has no `Communications` branch, so the request 200s with no effect — and even if a branch were added, `SetOrganization` now drops zero-value fields (see `dynamicUpdateDocument`), so `communications: false` could never persist. Worth either wiring it up (with a pointer/`*bool` in the request), or removing it from the docs/request surface in this same breaking change.

db/organizations.go
```diff
@@ -85,9 +85,7 @@ func (ms *MongoStorage) SetOrganization(org *Organization) error {
 	defer cancel()
 	// prepare the document to be updated in the database modifying only the
 	// fields that have changed
-	// define 'active' parameter to be updated always to update it even its new
-	// value is false
-	updateDoc, err := dynamicUpdateDocument(org, []string{"active"})
+	updateDoc, err := dynamicUpdateDocument(org, nil)
```
elboletaire, line 92: **[confirmed, design] Deleting the only org use of `alwaysUpdateTags` leaves no way to persist a zero value for any top-level Organization scalar.** `dynamicUpdateDocument` drops zero-value fields, so post-PR `Communications` can only ever go false→true and strings can never be cleared — the next toggleable bool added to `Organization` recreates the #625 bug class, this time as a silent no-op instead of a clobber (a future `if newOrgInfo.X != org.X` branch in the PUT handler would 200 while Mongo keeps the old value). Counters/subscription/meta have dedicated `$inc`/`$set` escape hatches, but the exact field class #625 belonged to (top-level scalars updated via PUT + `SetOrganization`) does not. A short comment here warning that zero values are not persisted — or moving the PUT body to pointer fields — would guard the recurrence.

~~Edit:~~ an earlier version of this comment also flagged a latent empty-`$set` upsert error for a bare-Address org; that was wrong — `db/org_members_test.go` exercises exactly that call (`SetOrganization(&Organization{Address: testOrgAddress})`) against mongo:7 and it succeeds, so the concern is withdrawn.

api/organizations_test.go
```diff
@@ -166,19 +166,15 @@ func TestUpdateOrganizationHandler(t *testing.T) {
...
-		Active: false,
-	}
-	resp, code = testRequest(t, http.MethodPut, token, activeUpdateInfo, "organizations", orgAddress.String())
-	c.Assert(code, qt.Equals, http.StatusOK, qt.Commentf("response: %s", resp))
-
-	// Verify active status update
+	// The organization payload must not expose an `active` flag (issue #625): nothing enforced
+	// it, and because it was a plain bool on a partial update, any request omitting it silently
+	// deactivated the organization.
 	resp, code = testRequest(t, http.MethodGet, token, nil, "organizations", orgAddress.String())
```
elboletaire: **[plausible, cleanup] Redundant second GET.** The organization was already fetched at line 142 and no state changes in between (the intervening PUTs all assert 4xx), so the `active`-absence check can reuse that response — or use the existing `requestAndParse[map[string]any]` helper from `api/api_test.go` — instead of a second round-trip against the containerized stack.

migrations/0020_drop_organization_active.go
```diff
@@ -0,0 +1,44 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.vocdoni.io/dvote/log"
+)
+
+func init() {
+	AddMigration(19, "drop_organization_active", upDropOrganizationActive, downDropOrganizationActive)
+}
+
+// upDropOrganizationActive removes the organizations' `active` field (issue #625). Nothing ever
```
elboletaire, line 16: **[plausible, cleanup] The #625 bug narrative is written out three times** — here, in `api/organizations_test.go:169-171`, and in `db/migration_drop_org_active_test.go`. Keeping the full rationale once (here) and having the tests carry a one-liner referencing issue #625 avoids three copies drifting independently — especially relevant given the premise correction suggested above.

db/migration_drop_org_active_test.go
```diff
@@ -0,0 +1,43 @@
...
+	database := testDB.DBClient.Database(testDB.database)
+
+	c.Assert(testDB.SetOrganization(&Organization{Address: testOrgAddress, CreatedAt: time.Now()}), qt.IsNil)
+	// Seed the field the way a database written before this migration holds it: the model no
+	// longer carries it, so it has to be set directly.
+	_, err := testDB.organizations.UpdateOne(ctx,
+		bson.M{"_id": testOrgAddress}, bson.M{"$set": bson.M{"active": false}})
+	c.Assert(err, qt.IsNil)
+
+	mig, ok := migrations.AsMap()[19]
```
elboletaire: **[plausible, cleanup] Third hand-rolled copy of the migration-registry lookup.** `db/migration_normalize_emails_test.go:16` (`runNormalizeEmailsMigration`) and `db/migration_question_slots_test.go:17` (`questionSlotsMigration`) each inline the same `migrations.AsMap()[N]` + assert pattern in this same package. Low stakes, but this PR is the point where a shared `migrationByVersion(c, n)` helper would pay for itself.

elboletaire (changes requested):

High-effort recall-biased review (8 finder angles, per-candidate verification). The removal itself is clean: nothing left reads the org-level flag, both create-request wrappers lose it via the embedded struct, migration numbering is correct, and the tree builds and vets clean. Seven findings survived verification — 3 confirmed, 4 plausible — posted inline. The two I would fix before merge are the migration's factual premise (the old PUT handler *did* honor a deliberate `active: false`, so the `$unset` is lossier than the godoc claims) and the `communications` field that the PUT docs still advertise as changeable while the handler ignores it — this PR edited both of those exact blocks. The rest are latent edges and cleanup.

lucasmenendez:
All seven findings addressed in 98b6eee.

**Confirmed**

1. *Migration premise* — you're right, the deleted `if newOrgInfo.Active != org.Active` branch honored a deliberate `PUT {"active": false}`. The godoc no longer claims every stored `false` was the bug; it says deactivation was possible but a stored `false` doesn't prove intent, since the same branch clobbered on an omitted field, and the two are indistinguishable. Rather than archive, the up now `Distinct`s `_id` on `{active: false}` and logs the addresses before the `$unset`, so the discarded state leaves a record.
2. *`communications` advertised but ignored* — resolved docs-only: dropped from the *Update organization* example. Wiring it up needs either `alwaysUpdateTags` (the exact clobber pattern this PR removes) or a new dedicated setter — new surface inside a breaking change. Also fixed the key in the *Create organization* example, which said `communication` where the field is really `communications` and *is* honored.
3. *Zero-value trap* — comment added above `dynamicUpdateDocument(org, nil)` naming the ceiling and the upgrade path (dedicated `$set` + pointer request field, not `alwaysUpdateTags`).

**Plausible**

4. *Down filter* — now `UpdateMany(ctx, bson.M{}, {$set: {active: true}})`, matching the stated policy. The migration test seeds a stray `false` post-rollback and asserts it is overwritten.
5. *Redundant GET* — dropped; the `active`-absence check reuses the line-142 response.
6. *Triplicated narrative* — full rationale kept only in the migration godoc; both test comments are one-liners referencing #625.
7. *Registry lookup* — `migrationByVersion(c, n)` in `db/migrations_test.go`; all three call sites use it and the two local helpers are gone.

`go build`/`go vet`/`check-qt-patterns.sh` clean, `golangci-lint run ./db/... ./api/... ./migrations/...` 0 issues, `./db/` and `./api/` migration + organization tests pass.

The `communications` PUT gap and the zero-value trap are the same latent class as #625 — I'll open a follow-up issue for both rather than grow this PR.

lucasmenendez:
Merged main and pushed af3aa83 — the lint failure was a merge artifact, not a lint rule:

- main's `db/census_revocation_test.go` and `db/org_members_test.go` build `Organization` literals with `Active: true`, which stops compiling once this branch removes the field (golangci-lint reports it as `typecheck`). Dropped from the four literals.
- main also landed `migrations/0019_census_revocation_indexes.go`, colliding with this branch's version 19. Renumbered mine to **0020**`_drop_organization_active.go`; the test now asks for `migrationByVersion(c, 20)`.

`go build`, `go vet`, `golangci-lint run ./db/... ./api/... ./migrations/...` and `check-qt-patterns.sh` clean on the merge result; `./db/` tests pass (34).

All seven review threads resolved.

elboletaire (approved):
Opus 5 here, controlled by elboletaire.

Reviewed the full diff against `main`, read the surrounding code, and ran the affected suites locally against real containers: `go build`/`go vet` clean, the whole `./db/` package passes, and `TestUpdateOrganizationHandler|TestOrganizationPartialUpdate|TestOrganizationWithOptionalFields` in `./api/` pass. Approving.

## Verified

- **No remaining reader.** Every surviving `Active` in the tree is `OrganizationSubscription.Active` / `SubscriptionDetails.Active` (`db/types.go`, `api/apicommon/types.go`), which the diff doesn't touch. `go vet ./...` builds the test files too, so there's no orphaned struct literal anywhere.
- **`dynamicUpdateDocument(org, nil)`.** `"active"` was the only entry in `alwaysUpdateTags`, and `nil` matches every other call site. The edge case worth checking is `$set` going empty — `SetOrganization(&Organization{Address: x})` in `db/organizations_test.go` sets only the `_id`-tagged field. Mongo 7 accepts an empty `$set` with upsert and that subtest passes, so it's not a regression.
- **The `$unset` is safe against schema validation.** No collection validator or index references `active` (the validators in `0001_initial_collections.go` cover only `users`, `subscriptions`, `organizationInvites`), and the driver ignores the leftover field when decoding already-written docs, so old binaries and new documents coexist fine mid-deploy.
- **Migration numbering.** `AddMigration(20, ...)` is unique after the renumber, and `RunMigrationsDown(all)` → `Up` round-trips (`TestMigrations` passes).
- **Docs.** Every remaining `"active"` in `docs/swagger.yaml` and `api/docs.md` is the subscription one. The `communication` → `communications` fix matches the actual JSON tag.
- **Behavioural delta in `updateOrganizationHandler`.** Dropping the `Active` branch means a PUT carrying no updatable field now skips the DB write entirely, where before `newOrgInfo.Active != org.Active` made nearly every PUT write. The only side effect skipped is the `addOrganizationToUser` re-link inside `SetOrganization`, which the handler's own admin check already guarantees — nothing can depend on it.

## Non-blocking

**`migrations/0020_drop_organization_active.go:25`** — the up migration `Distinct`s every `_id` with `active: false` and logs the whole slice in one line. By this PR's own analysis that value was written by *any* PUT omitting `active`, so in production the set is plausibly most organizations rather than a handful. That gives three problems: the startup log becomes a single entry with thousands of `primitive.Binary` addresses (unreadable, and heavy on log ingestion); past roughly 450k organizations the `Distinct` result exceeds the 16MB BSON limit, errors, and fails `RunMigrationsUp`, blocking the deploy; and line 39 logs `res.ModifiedCount` under the same `organizations` key as the array, so one field is an array in one log line and an int in the next. Logging `len(deactivated)` under a distinct key fixes all three.

**PR description drift** — the body still refers to `0019_drop_organization_active`; the migration was renumbered to `0020` in af3aa83.

## Coordination

The breaking change is real and the description is right that `@vocdoni/api-types` must drop the field. Tracked now in vocdoni/integrator-sdk#33 (`Organization.active` and `CreateOrganizationRequest.active`; `SubscriptionDetails.active` stays). That release coordination is the actual risk here — the Go side is sound.

PR #634, opened 2026-08-03
feat(processes): sign a batch of ballots at POST /processes/{processId}/sign-batch

lucasmenendez:
## Why

A multi-question voting process is **one on-chain election per question**, so a voter holds one ballot per question. `POST /votes` already relays them in a single call — the signing side had no counterpart. A voter had to hit `POST /processes/{processId}/sign` once per question, and each call re-did the same work (load the process, load and verify the auth token, load the org member, load the census, compute the weight) before signing one ballot.

## What

`POST /processes/{processId}/sign-batch` — one auth token, N ballots, one response.

```jsonc
// request
{"authToken": "…", "ballots": [
  {"upstreamId": "aa…", "address": "…"},
  {"upstreamId": "bb…", "address": "…"}
]}

// response
{"signatures": [
  {"upstreamId": "aa…", "signature": "…", "weight": "01"},
  {"upstreamId": "bb…", "error": "process already consumed"}
]}
```

Fields follow the `/processes` naming (`upstreamId`, `address`) that `ProcessQuestionStatus`, `QuestionConsumedAddress` and `db.VotingProcessQuestion` already use — not the `electionId`/`payload` of the shared/legacy `SignRequest` — so the ids read the same as in the `check` and `sign-info` responses a voter already called.

`ProcessSignHandler` is split into two reusable halves so the batch handler shares the exact checks rather than duplicating them:

- `resolveSignContext` — the per-request part: voting process, verified token anchored to it, member behind it, census weight.
- `authorizeQuestion` — the per-ballot part: the election belongs to this process, and the member is eligible for that question.

The single endpoint keeps its authorization semantics, with two review-driven changes: failures are sanitized through the same `signOutcome` mapping as the batch (no raw signer detail in the body), and an *internal* signer/storage failure is now a 500 rather than a 401 — authorization rejections stay 401.

## Failure semantics

Signing consumes a per-election slot that cannot be given back (`csp.Sign` → `db.ConsumeCSPProcess`), so the batch is **authorized as a unit and signs nothing on failure**:

| Condition | Result |
|---|---|
| election not in this process / member not eligible | 401, nothing signed |
| missing voter address | 400, nothing signed |
| repeated `upstreamId` | 400, nothing signed |
| empty batch / more ballots than the process has questions | 400, nothing signed |

Once authorized, every ballot is signed and the response carries one entry per request item, in order — a signature and weight, or the reason that election could not be signed. Those residual failures are per-ballot by nature (the slot is spent, or a concurrent request holds its `(user, election)` lock), so a failed entry is retried by calling again rather than rolled back. `MaxVoteOverwritesPerProcess` (10) makes that retry viable.

## Notes

- **No new cap constant** — a voter has at most one ballot per question, so `len(vp.QuestionIDs)` is the real bound (`maxQuestionsPerProcess` = 100 caps it transitively).
- **No new error codes** — reuses `ErrVoteBatchEmpty`, `ErrVoteBatchTooLarge`, `ErrRequestBodyTooLarge` from the vote relay, with the same `at index %d` annotation.
- **Body is capped** before it is buffered, as on the other public endpoints.
- **Signed sequentially** — `db.ConsumeCSPProcess` takes the storage write lock, so a parallel fan-out inside one request would serialize on it anyway.
- The legacy `POST /process/bundle/{bundleId}/sign` keeps its single-ballot form, but inherits two hardenings through the shared helpers: the 20-byte voter address check (`parseAddress` → `validateVoterAddress`) and the sanitized `signOutcome` error mapping.
- `CLAUDE.md` gains the two sections a new reader needs to reconstruct this from scratch: the two generations of the process API (and why several endpoints exist only in batch form), and the async transaction job queue behind every `202 + jobId`.

## Testing

`TestProcessCSPSignBatch` (`api/processes_csp_test.go`) covers: whole-batch rejection leaves nothing consumed, duplicate election → 400, empty → 400, over-cap → `ErrVoteBatchTooLarge`, missing address → 400, happy path signs both questions and `sign-info` reports both consumed, and a spent signing slot surfaces as a per-ballot `error` while its sibling still signs.

- `go test ./csp/... ./api/` → 294 passed
- `make lint` → 0 issues
- `scripts/check-qt-patterns.sh` → clean
- `docs/swagger.yaml` regenerated

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -248,7 +248,11 @@ const (
 	processesAuthEndpoint       = "/processes/{processId}/auth/{step}"
 	processesAuthResendEndpoint = "/processes/{processId}/auth/resend"
 	processesSignEndpoint       = "/processes/{processId}/sign"
-	processesWeightEndpoint     = "/processes/{processId}/weight"
+	// POST /processes/{processId}/sign-batch to sign every question's ballot in one call, so a
+	// voter casting a multi-question process does not need one round trip per question. The
+	// batch is authorized as a unit and signs nothing on failure.
+	processesSignBatchEndpoint = "/processes/{processId}/sign-batch"
+	processesWeightEndpoint    = "/processes/{processId}/weight"
 
 	// // census auth routes (currently not implemented)
 	// // POST /process/{processId}/auth/0 to initiate auth
--- a/api/routes.go
+++ b/api/routes.go
@@ -249,8 +249,9 @@ const (
 	processesAuthResendEndpoint = "/processes/{processId}/auth/resend"
 	processesSignEndpoint       = "/processes/{processId}/sign"
 	// POST /processes/{processId}/sign-batch to sign every question's ballot in one call, so a
-	// voter casting a multi-question process does not need one round trip per question. The
-	// batch is authorized as a unit and signs nothing on failure.
+	// voter casting a multi-question process does not need one round trip per question.
+	// Authorization is all or nothing — a rejected batch signs nothing — but the signing that
+	// follows reports per ballot: one entry can fail while its siblings succeed.
 	processesSignBatchEndpoint = "/processes/{processId}/sign-batch"
 	processesWeightEndpoint    = "/processes/{processId}/weight"
 
--- a/csp/handlers/processes.go
+++ b/csp/handlers/processes.go
@@ -11,12 +11,17 @@ import (
 	"github.com/go-chi/chi/v5"
 	"github.com/vocdoni/saas-backend/api/apicommon"
 	"github.com/vocdoni/saas-backend/csp"
+	"github.com/vocdoni/saas-backend/csp/signers"
 	"github.com/vocdoni/saas-backend/db"
 	"github.com/vocdoni/saas-backend/errors"
 	"github.com/vocdoni/saas-backend/internal"
 	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.vocdoni.io/dvote/log"
 	"go.vocdoni.io/dvote/vochain/state"
 )
+
+// apiErr lifts an API error to a pointer, so a helper can signal "no error" with nil.
+func apiErr(e errors.Error) *errors.Error { return &e }
 
 // parseProcessID parses the {processId} URL param (a voting-process Mongo ObjectID) and
 // returns both the ObjectID and its bytes, which are used as the CSP token anchor.
@@ -196,11 +201,7 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
 //	@Failure		500			{object}	errors.Error	"Internal server error"
 //	@Router			/processes/{processId}/sign [post]
 func (c *CSPHandlers) ProcessSignHandler(w http.ResponseWriter, r *http.Request) {
-	oid, anchor, ok := parseProcessID(w, r)
-	if !ok {
-		return
-	}
-	vp, ok := c.getVotingProcess(w, oid)
+	oid, _, ok := parseProcessID(w, r)
 	if !ok {
 		return
 	}
@@ -208,56 +209,233 @@ func (c *CSPHandlers) ProcessSignHandler(w http.ResponseWriter, r *http.Request)
 	if !ok {
 		return
 	}
-	auth, ok := c.getAuthInfo(w, req.AuthToken)
+	sc, ok := c.resolveSignContext(w, oid, req.AuthToken)
 	if !ok {
 		return
 	}
-	if !auth.Verified {
-		errors.ErrUnauthorized.WithErr(csp.ErrAuthTokenNotVerified).Write(w)
+	upstreamID, sErr := c.authorizeQuestion(sc, req.ProcessID)
+	if sErr != nil {
+		sErr.Write(w)
 		return
-	}
-	if !bytes.Equal(anchor, auth.BundleID) {
-		errors.ErrUnauthorized.Withf("token does not belong to the process").Write(w)
-		return
-	}
-	// resolve the target question by its on-chain election id and verify it belongs to
-	// this process
-	question, err := c.mainDB.QuestionByUpstreamID(req.ProcessID)
-	if err != nil && err != db.ErrNotFound {
-		errors.ErrGenericInternalServerError.WithErr(err).Write(w)
-		return
-	}
-	if err != nil || question.ProcessID != oid {
-		errors.ErrUnauthorized.Withf("election not found in process").Write(w)
-		return
-	}
-	// authorize the member against the question's eligibility subset
-	if !memberEligibleForQuestion(question, auth.UserID.String()) {
-		errors.ErrUnauthorized.Withf("member not eligible for this question").Write(w)
-		return
-	}
-	member, ok := c.orgMemberFromAuth(w, vp.OrgAddress, auth)
-	if !ok {
-		return
-	}
-	census, err := c.mainDB.Census(vp.CensusID.Hex())
-	if err != nil {
-		errors.ErrCensusNotFound.WithErr(err).Write(w)
-		return
-	}
-	weight := uint64(1)
-	if census.Weighted {
-		if member.Weight == 0 {
-			errors.ErrZeroWeightVoter.Write(w)
-			return
-		}
-		weight = member.Weight
 	}
 	address, ok := parseAddress(w, req.Payload)
 	if !ok {
 		return
 	}
-	c.signAndRespond(w, req.AuthToken, *address, question.UpstreamID, big.NewInt(int64(weight)).Bytes())
+	c.signAndRespond(w, req.AuthToken, *address, upstreamID, sc.weight)
+}
+
+// signContext is the part of a ballot signing request that is resolved once per call, from
+// the process named in the path and the auth token in the body: the voting process, the
+// member behind the token and their census weight. A batch signs every ballot under one of
+// these, which is the whole point of the batch endpoint.
+type signContext struct {
+	process  *db.VotingProcess
+	memberID string
+	weight   internal.HexBytes
+}
+
+// resolveSignContext runs the per-request half of a ballot signing request: load the voting
+// process, check the auth token is verified and anchored to it, resolve the org member behind
+// it and compute their census weight. It writes the proper error and returns false on failure.
+func (c *CSPHandlers) resolveSignContext(
+	w http.ResponseWriter, oid primitive.ObjectID, authToken internal.HexBytes,
+) (*signContext, bool) {
+	vp, ok := c.getVotingProcess(w, oid)
+	if !ok {
+		return nil, false
+	}
+	auth, ok := c.getAuthInfo(w, authToken)
+	if !ok {
+		return nil, false
+	}
+	if !auth.Verified {
+		errors.ErrUnauthorized.WithErr(csp.ErrAuthTokenNotVerified).Write(w)
+		return nil, false
+	}
+	if !bytes.Equal(oid[:], auth.BundleID) {
+		errors.ErrUnauthorized.Withf("token does not belong to the process").Write(w)
+		return nil, false
+	}
+	member, ok := c.orgMemberFromAuth(w, vp.OrgAddress, auth)
+	if !ok {
+		return nil, false
+	}
+	census, err := c.mainDB.Census(vp.CensusID.Hex())
+	if err != nil {
+		errors.ErrCensusNotFound.WithErr(err).Write(w)
+		return nil, false
+	}
+	weight := uint64(1)
+	if census.Weighted {
+		if member.Weight == 0 {
+			errors.ErrZeroWeightVoter.Write(w)
+			return nil, false
+		}
+		weight = member.Weight
+	}
+	return &signContext{
+		process:  vp,
+		memberID: auth.UserID.String(),
+		weight:   big.NewInt(int64(weight)).Bytes(),
+	}, true
+}
+
+// authorizeQuestion runs the per-ballot half of a signing request: resolve the target question
+// by its on-chain election id, verify it belongs to this process and authorize the member
+// against the question's eligibility subset. It returns the question's on-chain election id, or
+// the API error to write back, never both.
+func (c *CSPHandlers) authorizeQuestion(
+	sc *signContext, electionID internal.HexBytes,
+) (internal.HexBytes, *errors.Error) {
+	question, err := c.mainDB.QuestionByUpstreamID(electionID)
+	if err != nil && err != db.ErrNotFound {
+		return nil, apiErr(errors.ErrGenericInternalServerError.WithErr(err))
+	}
+	if err != nil || question.ProcessID != sc.process.ID {
+		return nil, apiErr(errors.ErrUnauthorized.Withf("election not found in process"))
+	}
+	if !memberEligibleForQuestion(question, sc.memberID) {
+		return nil, apiErr(errors.ErrUnauthorized.Withf("member not eligible for this question"))
+	}
+	return question.UpstreamID, nil
+}
+
+// maxSignBatchBodyBytes bounds a POST /processes/{processId}/sign-batch body. One ballot is an
+// election id plus a voter address, ~200 characters of JSON once hex-encoded and quoted, so 512
+// bytes each leaves ample headroom; a process holds at most 100 questions, and a voter at most
+// one ballot per question. Like the relay endpoints this one is public, so the body has to be
+// bounded before it is buffered.
+const maxSignBatchBodyBytes = 100*512 + 4<<10
+
+// ballot is one validated item of a sign batch: the on-chain election to sign for and the
+// voter address to sign, both resolved during the validation pass.
+type ballot struct {
+	upstreamID internal.HexBytes
+	address    internal.HexBytes
+}
+
+// ProcessSignBatchHandler godoc
+//
+//	@Summary		Sign every question's ballot of a voting process in one call
+//	@Description	Batch form of POST /processes/{processId}/sign. A multi-question voting process is
+//	@Description	one on-chain election per question, so a voter holds one ballot per question and
+//	@Description	signing them one by one costs a round trip each. This signs them all under a single
+//	@Description	verified auth token. Public endpoint: the token authenticates the voter.
+//	@Description	The batch is authorized as a unit and signs nothing on failure — every ballot must
+//	@Description	name an election of this process (else 401) the member is eligible for (else 401),
+//	@Description	carry a parseable voter address (else 400), and no election may be repeated (else
+//	@Description	400). Once authorized, every ballot is signed and the response carries one entry per
+//	@Description	request item, in order: a signature and weight, or the reason that election could not
+//	@Description	be signed (its signing slot is spent, or a concurrent request holds it). Signing
+//	@Description	consumes a per-election slot that cannot be given back, so a failed entry is retried
+//	@Description	by calling again, not rolled back.
+//	@Tags			processes
+//	@Accept			json
+//	@Produce		json
+//	@Param			processId	path		string						true	"Process ID"
+//	@Param			request		body		handlers.SignBatchRequest	true	"Auth token and one ballot per question"
+//	@Success		200			{object}	handlers.SignBatchResponse
+//	@Failure		400			{object}	errors.Error	"Invalid input data"
+//	@Failure		401			{object}	errors.Error	"Unauthorized, unverified token, election not in process, or member not eligible"
+//	@Failure		404			{object}	errors.Error	"Process, census, or user not found"
+//	@Failure		413			{object}	errors.Error	"Request body too large"
+//	@Failure		500			{object}	errors.Error	"Internal server error"
+//	@Router			/processes/{processId}/sign-batch [post]
+//
+//nolint:lll
+func (c *CSPHandlers) ProcessSignBatchHandler(w http.ResponseWriter, r *http.Request) {
+	oid, _, ok := parseProcessID(w, r)
+	if !ok {
+		return
+	}
+	req, ok := parseSignBatchRequest(w, r)
+	if !ok {
+		return
+	}
+	sc, ok := c.resolveSignContext(w, oid, req.AuthToken)
+	if !ok {
+		return
+	}
+	switch {
+	case len(req.Signatures) == 0:
+		errors.ErrVoteBatchEmpty.Write(w)
+		return
+	case len(req.Signatures) > len(sc.process.QuestionIDs):
+		// a voter holds at most one ballot per question, so the process itself is the cap.
+		errors.ErrVoteBatchTooLarge.Withf("%d ballots, the process has %d questions",
+			len(req.Signatures), len(sc.process.QuestionIDs)).Write(w)
+		return
+	}
+
+	// authorize the whole batch before signing anything: signing consumes a per-election slot
+	// that cannot be given back, so a rejected batch must leave the voter with nothing signed
+	// rather than with a signed prefix.
+	ballots := make([]ballot, len(req.Signatures))
+	seen := make(map[string]int, len(req.Signatures))
+	for i, item := range req.Signatures {
+		// a repeated election would contend with itself on the per-(user, election) signer
+		// lock, and a voter has at most one ballot per question anyway.
+		if first, dup := seen[string(item.ProcessID)]; dup {
+			errors.ErrMalformedBody.Withf(
+				"the ballot at index %d repeats the election of index %d", i, first).Write(w)
+			return
+		}
+		seen[string(item.ProcessID)] = i
+		upstreamID, sErr := c.authorizeQuestion(sc, item.ProcessID)
+		if sErr != nil {
+			sErr.Withf("at index %d", i).Write(w)
+			return
+		}
+		address := new(internal.HexBytes)
+		if err := address.ParseString(item.Payload); err != nil {
+			errors.ErrMalformedBody.WithErr(err).Withf("at index %d", i).Write(w)
+			return
+		}
+		ballots[i] = ballot{upstreamID: upstreamID, address: *address}
+	}
+
+	// ponytail: signed sequentially. db.ConsumeCSPProcess takes the storage write lock, so a
+	// parallel fan-out inside one request would serialize on it anyway; revisit only if that
+	// lock stops being the bottleneck.
+	resp := &SignBatchResponse{Signatures: make([]SignBatchResult, len(ballots))}
+	for i, b := range ballots {
+		res := &resp.Signatures[i]
+		res.ProcessID = b.upstreamID
+		signature, err := c.csp.Sign(req.AuthToken, b.address, b.upstreamID, sc.weight,
+			signers.SignerTypeECDSASalted)
+		if err != nil {
+			// per-ballot by nature: this election's signing slot is spent, or a concurrent
+			// request holds its lock. The batch's siblings are unaffected.
+			log.Warnw("could not sign a batch ballot", "procId", b.upstreamID, "error", err)
+			res.Error = err.Error()
+			continue
+		}
+		res.Signature = signature
+		res.Weight = sc.weight
+	}
+	apicommon.HTTPWriteJSON(w, resp)
+}
+
+// parseSignBatchRequest decodes a capped sign-batch body and checks it carries an auth token,
+// writing the proper error and returning false on failure.
+func parseSignBatchRequest(w http.ResponseWriter, r *http.Request) (*SignBatchRequest, bool) {
+	r.Body = http.MaxBytesReader(w, r.Body, maxSignBatchBodyBytes)
+	var req SignBatchRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		var tooLarge *http.MaxBytesError
+		if errors.As(err, &tooLarge) {
+			errors.ErrRequestBodyTooLarge.Withf("the limit is %d bytes", maxSignBatchBodyBytes).Write(w)
+			return nil, false
+		}
+		errors.ErrMalformedBody.Write(w)
+		return nil, false
+	}
+	if req.AuthToken == nil {
+		errors.ErrUnauthorized.Withf("missing auth token").Write(w)
+		return nil, false
+	}
+	return &req, true
 }
 
 // ProcessWeightHandler godoc
--- a/csp/handlers/processes.go
+++ b/csp/handlers/processes.go
@@ -19,11 +19,6 @@ import (
 	"go.vocdoni.io/dvote/vochain/state"
 )
 
-// apiErr lifts an API error to a pointer, so a helper can signal "no error" with nil. Named
-// asErr (not apiErr) because handlers.go has several `apiErr, ok := err.(errors.Error)` locals
-// that would shadow a same-named helper.
-func asErr(e errors.Error) *errors.Error { return &e }
-
 // parseProcessID parses the {processId} URL param (a voting-process Mongo ObjectID) and
 // returns both the ObjectID and its bytes, which are used as the CSP token anchor.
 func parseProcessID(w http.ResponseWriter, r *http.Request) (primitive.ObjectID, internal.HexBytes, bool) {
@@ -285,7 +280,8 @@ func (c *CSPHandlers) resolveSignContext(
 // authorizeQuestion runs the per-ballot half of a signing request: resolve the target question
 // by its on-chain election id, verify it belongs to this process and authorize the member
 // against the question's eligibility subset. It returns the question's on-chain election id, or
-// the API error to write back, never both.
+// the API error to write back, never both. The batch handler resolves against a preloaded map
+// instead (authorizeLoadedQuestion) to avoid one query per ballot.
 func (c *CSPHandlers) authorizeQuestion(
 	sc *signContext, electionID internal.HexBytes,
 ) (internal.HexBytes, *errors.Error) {
@@ -293,17 +289,30 @@ func (c *CSPHandlers) authorizeQuestion(
 	// ErrInvalidData, which the branch below would surface to the client as a 500. The message is
 	// field-neutral because the single-sign endpoint sends this same id as "electionId".
 	if len(electionID) == 0 {
-		return nil, asErr(errors.ErrMalformedBody.With("missing election id"))
+		return nil, errors.Ptr(errors.ErrMalformedBody.With("missing election id"))
 	}
 	question, err := c.mainDB.QuestionByUpstreamID(electionID)
 	if err != nil && !errors.Is(err, db.ErrNotFound) {
-		return nil, asErr(errors.ErrGenericInternalServerError.WithErr(err))
+		return nil, errors.Ptr(errors.ErrGenericInternalServerError.WithErr(err))
 	}
-	if err != nil || question.ProcessID != sc.process.ID {
-		return nil, asErr(errors.ErrUnauthorized.Withf("election not found in process"))
+	if err != nil {
+		question = nil // not found: authorizeLoadedQuestion folds it into the same 401
+	}
+	return authorizeLoadedQuestion(sc, question)
+}
+
+// authorizeLoadedQuestion authorizes one already-loaded question (nil means the election id
+// resolved to nothing): it must belong to this process and the member must be in its
+// eligibility subset. Not-found and wrong-process collapse into the same 401 on purpose — a
+// caller learns nothing about elections outside the process it is signing for.
+func authorizeLoadedQuestion(
+	sc *signContext, question *db.VotingProcessQuestion,
+) (internal.HexBytes, *errors.Error) {
+	if question == nil || question.ProcessID != sc.process.ID {
+		return nil, errors.Ptr(errors.ErrUnauthorized.Withf("election not found in process"))
 	}
 	if !memberEligibleForQuestion(question, sc.memberID) {
-		return nil, asErr(errors.ErrUnauthorized.Withf("member not eligible for this question"))
+		return nil, errors.Ptr(errors.ErrUnauthorized.Withf("member not eligible for this question"))
 	}
 	return question.UpstreamID, nil
 }
@@ -339,10 +348,14 @@ type authorizedBallot struct {
 //	@Description	Once authorized, every ballot is signed and the response is always 200 with one
 //	@Description	entry per request item, in order — even if every ballot fails (the request was
 //	@Description	honored; the outcome is per item). Each entry carries a signature and weight, or a
-//	@Description	stable machine-readable code plus a message. Retry ONLY the entries that carry a
-//	@Description	code: re-sending the whole batch re-signs the ballots that already succeeded, and
-//	@Description	each re-sign counts toward the election's finite overwrite budget
-//	@Description	(MaxVoteOverwritesPerProcess, 10) — past it the election is permanently locked.
+//	@Description	stable machine-readable code plus a message: already_consumed (that election's
+//	@Description	signing slot is spent; not retryable), already_signing (a concurrent request holds
+//	@Description	it; retry), auth_invalid (the token was invalidated mid-batch and every remaining
+//	@Description	entry is stamped with it; authenticate again) or sign_failed (internal failure;
+//	@Description	retry). Retry ONLY the entries that carry a retryable code: re-sending the whole
+//	@Description	batch re-signs the ballots that already succeeded, and each re-sign counts toward
+//	@Description	the election's finite overwrite budget (MaxVoteOverwritesPerProcess, 10) — past it
+//	@Description	the election is permanently locked.
 //	@Tags			processes
 //	@Accept			json
 //	@Produce		json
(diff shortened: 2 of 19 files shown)


csp/handlers/processes.go
```diff
@@ -196,68 +201,241 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+		return nil, apiErr(errors.ErrGenericInternalServerError.WithErr(err))
+	}
+	if err != nil || question.ProcessID != sc.process.ID {
+		return nil, apiErr(errors.ErrUnauthorized.Withf("election not found in process"))
+	}
+	if !memberEligibleForQuestion(question, sc.memberID) {
+		return nil, apiErr(errors.ErrUnauthorized.Withf("member not eligible for this question"))
+	}
+	return question.UpstreamID, nil
+}
```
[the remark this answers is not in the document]
lucasmenendez: Good catch, and it was worse than the batch path: `authorizeQuestion` is shared with `POST /processes/{processId}/sign`, so the false 500 was reachable there too (and was pre-existing before this PR).

Fixed in 09de86b by guarding at the top of the helper rather than mapping `ErrInvalidData` afterwards — `db.QuestionByUpstreamID` only returns it for `len(upstreamID) == 0`, so excluding that case up front makes the branch unreachable instead of leaving two ways to reach the same outcome:

```go
if len(electionID) == 0 {
    return nil, apiErr(errors.ErrMalformedBody.With("missing upstreamId"))
}
```

In the batch handler it surfaces as `at index N` like the other per-ballot rejections and still rejects the whole batch. Covered by new assertions in both `TestProcessCSP` and `TestProcessCSPSignBatch`.

csp/handlers/processes.go
```diff
@@ -196,68 +201,241 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+			return nil, false
 		}
 		weight = member.Weight
 	}
-	address, ok := parseAddress(w, req.Payload)
+	return &signContext{
+		process:  vp,
+		memberID: auth.UserID.String(),
+		weight:   big.NewInt(int64(weight)).Bytes(),
+	}, true
```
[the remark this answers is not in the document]
lucasmenendez: Real, though only above 2^63 — `db.OrgMember.Weight` is a `uint64` and `int64(weight)` wraps negative there.

The same conversion appeared at six call sites (three in `processes.go`, three in the deprecated bundle handlers), so fixing it only here would have left five copies of the bug. 09de86b replaces all six with one helper:

```go
func weightBytes(w uint64) internal.HexBytes {
    return new(big.Int).SetUint64(w).Bytes()
}
```

Net -5 lines and `math/big` drops out of `processes.go` entirely.

csp/handlers/processes.go
```diff
@@ -196,68 +201,241 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+		res.ProcessID = b.upstreamID
+		signature, err := c.csp.Sign(req.AuthToken, b.address, b.upstreamID, sc.weight,
+			signers.SignerTypeECDSASalted)
+		if err != nil {
+			// per-ballot by nature: this election's signing slot is spent, or a concurrent
+			// request holds its lock. The batch's siblings are unaffected.
+			log.Warnw("could not sign a batch ballot", "procId", b.upstreamID, "error", err)
+			res.Error = err.Error()
+			continue
+		}
```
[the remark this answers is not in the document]
lucasmenendez: Agreed — a voter retrying, or two tabs racing on the same question, would have produced a warn per occurrence and buried the failures that actually need attention. Fixed in 09de86b:

```go
func logSignFailure(upstreamID internal.HexBytes, err error) {
    if errors.Is(err, csp.ErrProcessAlreadyConsumed) || errors.Is(err, csp.ErrUserAlreadySigning) {
        log.Debugw("skipped a batch ballot", "procId", upstreamID, "reason", err)
        return
    }
    log.Warnw("could not sign a batch ballot", "procId", upstreamID, "error", err)
}
```

Kept at debug rather than dropped: the caller is told about these in the response body either way, so the log is only there for support. Warn stays for the unexpected ones (`ErrSign`, storage failures).

csp/handlers/types.go
```diff
@@ -57,6 +57,34 @@ type SignRequest struct {
...
+	AuthToken  internal.HexBytes `json:"authToken" swaggertype:"string" format:"hex" example:"deadbeef"`
+	Signatures []SignBatchItem   `json:"signatures"`
+}
+
+// SignBatchItem is one ballot of a SignBatchRequest: the question's on-chain election id and
+// the voter address to sign, matching SignRequest's electionId and payload fields.
+type SignBatchItem struct {
+	ProcessID internal.HexBytes `json:"electionId" swaggertype:"string" format:"hex" example:"deadbeef"`
+	Payload   string            `json:"payload"`
+}
```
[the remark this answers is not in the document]
lucasmenendez: Same conclusion, landed in b7b9e3a (before this review came in) — but renamed to `UpstreamID`/`json:"upstreamId"` rather than `ElectionID`/`electionId`.

Reason: `electionId` is the *legacy* spelling. It survives on `SignRequest` and `CheckMembershipRequest`, which are shared with the deprecated bundle handlers. The `/processes` generation consistently calls this `upstreamId` — see `ProcessQuestionStatus`, `QuestionConsumedAddress` and `db.VotingProcessQuestion` — so a voter reads the same key here as in the `check` and `sign-info` responses they already called to obtain the id.

`payload` got the same treatment (`address`, now `internal.HexBytes` like `QuestionConsumedAddress.Address`), and the request list is `ballots`, since they only become signatures on the way back.

api/routes.go
```diff
@@ -248,7 +248,11 @@ const (
 	processesAuthEndpoint       = "/processes/{processId}/auth/{step}"
 	processesAuthResendEndpoint = "/processes/{processId}/auth/resend"
 	processesSignEndpoint       = "/processes/{processId}/sign"
-	processesWeightEndpoint     = "/processes/{processId}/weight"
+	// POST /processes/{processId}/sign-batch to sign every question's ballot in one call, so a
+	// voter casting a multi-question process does not need one round trip per question. The
+	// batch is authorized as a unit and signs nothing on failure.
```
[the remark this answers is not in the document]
lucasmenendez: Fair — "signs nothing on failure" read as covering the signing phase too. Reworded in 09de86b so the scope of each is explicit:

```go
// Authorization is all or nothing — a rejected batch signs nothing — but the signing that
// follows reports per ballot: one entry can fail while its siblings succeed.
```

elboletaire (changes requested):
This is Fable speaking on behalf of elboletaire. The findings below are the synthesis of a two-model cross-review (Opus and Fable reviewed the diff independently, then each verified the other's findings against the code); everything listed survived that verification pass.

## 🔴 Fix before merge

**1. Batch cap on `len(QuestionIDs)` bricks sign-batch for processes predating the field** — `csp/handlers/processes.go:370-374`
`QuestionIDs` is only written by `writeQuestions` on draft create/update, and no migration backfills it (`api/processes.go:1023` and migration 0018 both explicitly treat empty as "predates the field"). For any such process every sign-batch call — even a single ballot — is rejected with `400 ErrVoteBatchTooLarge` ("1 ballots, the process has 0 questions"), while the single-sign path keeps working. Unrecoverable for the client and a misleading message. Cap with `maxQuestionsPerProcess` (100) instead, or use `len(QuestionIDs)` only when non-zero and fall back to `QuestionsByProcess(oid)`.

## 🟠 Should fix

**2. Per-ballot errors carry no machine-readable code; an all-failed batch still returns 200** — `ProcessSignBatchHandler`
`SignBatchResult.Error` is a free-form string, so a client cannot distinguish terminal (`ErrProcessAlreadyConsumed`) from retryable (`ErrUserAlreadySigning`) without substring matching — unlike every other response in the API, where `errors.Error` carries a `code`. And since the handler always ends in `HTTPWriteJSON`, a batch where *every* ballot failed is still a 200. Add a stable per-result code (reuse the `errors.Error` codes) and document the all-failed-still-200 contract explicitly.

**3. Raw `err.Error()` echoed on a public endpoint** — `csp/handlers/processes.go:418`
`csp.Sign` returns `errors.Join(ErrSign, err)` for signer failures (`csp/sign.go:33`), so internal ECDSA error text reaches the voter-facing response body. To be fair this is parity with the single-sign path (which already echoes via `ErrUnauthorized.WithErr(err)`), not a regression — but the batch is the right moment to map known outcomes to sanitized, enumerated reasons and keep the raw detail in `logSignFailure` only.

**4. The retry wording invites burning the overwrite budget** — swagger/godoc text + `db/csp.go:241-251`
The docs say "a failed entry is retried by calling again". A client that re-sends the *whole* batch re-signs the ballots that already succeeded, and each re-sign increments `TimesVoted`; after `MaxVoteOverwritesPerProcess` (10) those elections become permanently `ErrProcessAlreadyConsumed`. Say explicitly "retry only the entries that carry an error" and mention the finite overwrite budget.

**5. Voter address is never length-validated** — `SignBatchBallot.Address` (and pre-existing `parseAddress`)
The batch only rejects an *empty* address; `internal.HexBytes` accepts any length, and `HexBytes.ParseString` even pads odd-length input. The address is signed into the CA bundle and pinned as `UsedAddress` (`db/csp.go:248`), which rejects every later overwrite whose address differs — so one wrong-length address locks the voter out of that election permanently. The gap is pre-existing on `/sign` too, but the batch's new validation pass is the natural place to add `len(item.Address) != common.AddressLength` (and a sane length bound for `UpstreamID`).

## 🟡 Nits

**6.** `authorizeQuestion`'s "missing upstreamId" error is also reached from the single-sign endpoint, whose JSON field is `electionId` — the error names a field that body doesn't have. Use a field-neutral message or pass the field name in from the caller.

**7.** `err != db.ErrNotFound` in `authorizeQuestion` — correct today (the sentinel is returned bare), but a future wrap silently turns a 401 into a 500. Convention is `errors.Is`, already used elsewhere in this file.

**8.** The new package-level `apiErr` is shadowed by same-named locals in this file and `handlers.go`, and duplicates `asPtr` in `api/process_vote.go`. Rename or hoist a single shared helper.

**9.** `parseSignBatchRequest` reimplements `decodeCappedJSON`'s MaxBytesReader + 413 mapping from `api/process_vote.go`. The duplication is currently forced by the import cycle (`api` imports `csp/handlers`), so the fix is hoisting the helper into `apicommon`.

**10.** `maxSignBatchBodyBytes` hardcodes `100`, silently duplicating the unexported `api.maxQuestionsPerProcess`. Same import-cycle constraint; at minimum name the source constant in the comment (per the `api/process_vote.go:88-104` precedent) so a future cap change greps its way here.

**11.** Tests: `c.Assert(bytes.Equal(...), qt.IsTrue)` (weight and the two `UpstreamID` comparisons) loses the failure diff. Use `qt.DeepEquals` / `qt.ContentEquals` — note the `internal.HexBytes` type conversion so `DeepEquals` matches.

**12.** Test coverage gaps: no 413 body-cap case, no unverified-token case, and no all-ballots-failed case (the failure case in the test is mixed: one ballot fails, one succeeds).

**13.** The `len(req.Ballots) == 0` check runs after `resolveSignContext`'s four DB lookups. Not a DoS lever (an invalid token dies at lookup 2 and the body is capped pre-decode), but the empty check needs nothing — move it first. The too-large check must stay where it is, it needs the resolved process.

## ℹ️ Info / follow-ups (not blocking this PR)

- `logSignFailure` logs only `procId`; a warn cannot be tied back to a voter during an incident even though `sc.memberID` is in scope. Add member id (and process oid) to the fields.
- The `//nolint:lll` sits inside the godoc block, so it suppresses `lll` for the entire function body and leaks into godoc output. Scope it to the offending line.
- The documented `404 "Process … not found"` is actually a 400 (`ErrMalformedURLParam` in `getVotingProcess`); the census/user halves of that 404 line are accurate. Inherited from `/sign`'s annotation.
- Pre-existing, worth tracking issues rather than fixes here: (a) `prepareSaltedKeySigner`'s error paths return a nil `userID`, so `Sign`'s deferred unlock hashes the wrong key and the (user, election) lock leaks until restart — practically unreachable from this endpoint but the batch contract leans on locks always releasing; (b) neither sign path re-checks `CensusParticipant` after auth, so a member removed from the census keeps signing until the token expires.

Also verified and worth stating: the batch-wide authorization strictly precedes the first `csp.Sign` call (a rejected batch consumes nothing), the duplicate check is byte-exact on decoded hex, the `weightBytes` fix correctly closes the int64 wrap at all call sites, and the body cap maps `*http.MaxBytesError` to 413 correctly. The overall shape — cap, resolve context once, authorize everything, then sign — matches the `/votes` batch precedent. Requesting changes solely for item 1 plus the should-fix group.

lucasmenendez:
Thanks — the cross-review earned its keep. I verified every item against the code before touching anything; all thirteen hold, and **item 1 was the real prize**. Everything below is in 355629d.

## 🔴 1 — batch cap on `len(QuestionIDs)`

Confirmed and fixed. I traced it exactly as described: migration 0018 is `unique_process_question_slots`, not a backfill, and `questionSetProblem` (`api/processes.go`) explicitly treats empty `QuestionIDs` as "predates the field" — so for those processes the endpoint rejected *every* batch, single ballot included, with a message that read like the client's fault. Straightforwardly unshippable.

Took the constant route rather than the `QuestionsByProcess` fallback: the fallback buys an extra DB read to reconstruct a number that was only ever a sanity ceiling — the duplicate-election check is what actually enforces one-ballot-per-question. Promoted `maxQuestionsPerProcess` to **`db.MaxQuestionsPerProcess`**, which also closes **item 10** (the hardcoded `100` in `maxSignBatchBodyBytes` now reads the same constant, no comment needed to keep them in sync).

Since the cap no longer needs the resolved process, **both** size checks moved above `resolveSignContext` — that's **item 13** for free.

## 🟠 2 + 3 — codes and sanitized messages

Both fixed together, since they're the same seam. `SignBatchResult` gains `code`, and a `signOutcome` helper maps the sentinels to `already_consumed` / `already_signing` / `sign_failed` plus a message safe for a public body. You were right that `errors.Join(ErrSign, err)` was leaking internal signer text — that now stays in `logSignFailure`.

Kept **all-failed-still-200**: the request *was* honored and the outcome is genuinely per item, which is the same contract `/votes/verify` uses for `verified=false`. The problem was that it was implied rather than stated, so the godoc now says it outright.

## 🟠 4 — retry wording

Fixed. "Retried by calling again" was actively harmful advice given a finite overwrite budget — the docs now say to retry **only the entries carrying a code**, and name `MaxVoteOverwritesPerProcess`.

## 🟠 5 — address length

Fixed in both places. `HexBytes` padding odd-length input into a silently-wrong address that then gets *pinned* as the election's consumer is a nasty failure mode — one bad value locks the voter out permanently. Added `common.AddressLength` in the batch pass **and** in `parseAddress`, so single `/sign` gets it too. Same reasoning as last round: it's a shared choke point, fixing one caller would leave the other broken.

## 🟡 Nits

All applied: field-neutral missing-id message (**6**), `errors.Is` (**7**), `apiErr` → `asErr` (**8** — you're right, `handlers.go` had six locals shadowing it), `decodeCappedJSON` hoisted to `apicommon.DecodeCappedJSON` (**9** — `errors` is a leaf package so the cycle isn't real, the duplication was just inherited), `qt.DeepEquals` (**11**), and the test gaps (**12**): 413 body cap, unverified token, all-failed batch, wrong-length address, and code assertions.

Also took the three cheap ℹ️ items: `logSignFailure` now logs member + process id, the stale `//nolint:lll` is gone (post-rename the longest godoc line is 122 < 130), and the `404` annotation no longer claims a not-found process, which is a 400.

## ℹ️ Deferred

Agreed these are pre-existing and out of scope here — filing them separately rather than smuggling them into this PR:

- `prepareSaltedKeySigner` returning a nil `userID` on its error paths, so `Sign`'s deferred unlock hashes the wrong key and leaks the (user, election) lock. Unreachable from this endpoint, but the batch contract does lean on locks always releasing.
- Neither sign path re-checking `CensusParticipant` after auth.

**Verification:** `go test ./csp/... ./api/ ./db/ ./api/apicommon/` → 539 passed. `make lint` → 0 issues. qt patterns clean. Swagger regenerated.

csp/handlers/processes.go
```diff
@@ -196,68 +202,285 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+
+// signOutcome maps a per-ballot signing error to a stable, machine-readable code and a message
+// safe for the public response body. The raw error (which for signer failures is
+// errors.Join(ErrSign, <internal detail>)) is kept out of the response and logged instead.
+func signOutcome(err error) (code, message string) {
+	switch {
+	case errors.Is(err, csp.ErrProcessAlreadyConsumed):
+		return "already_consumed", "this election's signing slot is already consumed"
+	case errors.Is(err, csp.ErrUserAlreadySigning):
+		return "already_signing", "a concurrent request is signing this election"
```
elboletaire: **Pre-existing lock leak that this retry contract turns into a permanent lockout.** In `csp/sign.go`, `prepareSaltedKeySigner` takes `c.lock(authTokenData.UserID, processID)` (line 79) and can then return with a **nil userID** on the `!authTokenData.Verified` branch (lines 81-83, reachable for a first-time signer whose token exists but was never verified: `IsCSPProcessConsumed` finds no consume record and returns `(false, nil)`, bypassing the earlier `ErrTokenNotVerified` switch). `Sign`'s `defer c.unlock(userID, processID)` then hashes `sha256(nil||processID)` — a different key than the one locked — so the `(user, election)` lock is never released. Every later attempt returns `ErrUserAlreadySigning` until restart, which this endpoint reports as a stable `already_signing` code whose documented guidance (retry) can never succeed. Fix in `prepareSaltedKeySigner`: unlock before those error returns (or return the userID alongside the error so the defer unlocks the right key).

csp/handlers/processes.go
```diff
@@ -196,68 +202,285 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+		res.Signature = signature
+		res.Weight = sc.weight
+	}
+	apicommon.HTTPWriteJSON(w, resp)
+}
+
+// signOutcome maps a per-ballot signing error to a stable, machine-readable code and a message
+// safe for the public response body. The raw error (which for signer failures is
+// errors.Join(ErrSign, <internal detail>)) is kept out of the response and logged instead.
+func signOutcome(err error) (code, message string) {
```
elboletaire, line 542: `signOutcome` collapses `csp.ErrInvalidAuthToken` / `csp.ErrAuthTokenNotVerified` (and transient Mongo failures) into the catch-all `sign_failed` inside a **200**. If the token is deleted mid-batch (`DeleteCSPAuthByBundle` via managed-access removal or user erasure) or Mongo blips, every remaining ballot comes back `sign_failed`, and the documented contract — "retry ONLY the entries that carry a code" — sends the client into futile retries instead of re-authenticating. The loop also keeps issuing one doomed `Sign` (DB round trips included) per remaining ballot. Suggest a distinct `auth_invalid` code (or aborting the batch with 401, matching what `resolveSignContext` would have returned), and short-circuiting the loop on `ErrInvalidAuthToken` since it dooms the whole batch.

csp/handlers/processes.go
```diff
@@ -196,68 +202,285 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+	if !ok {
 		return
 	}
-	// resolve the target question by its on-chain election id and verify it belongs to
-	// this process
-	question, err := c.mainDB.QuestionByUpstreamID(req.ProcessID)
-	if err != nil && err != db.ErrNotFound {
-		errors.ErrGenericInternalServerError.WithErr(err).Write(w)
-		return
+	c.signAndRespond(w, req.AuthToken, *address, upstreamID, sc.weight)
```
elboletaire, line 250: Inconsistent with the batch path this PR adds: `signAndRespond` (handlers.go:352-361) maps **every** `csp.Sign` failure to `errors.ErrUnauthorized.WithErr(err)`, and `Error.MarshalJSON` serializes the joined error text verbatim — so the same signer/storage failure that `/sign-batch` sanitizes to `sign_failed` ("the raw error stays in the log") leaks the raw `errors.Join(ErrSign, <internal detail>)` string to the client on `/processes/{id}/sign`, as a 401 rather than a 500. Since the single-sign path now shares `resolveSignContext`/`authorizeQuestion` with the batch, it seems natural to route its outcome through `signOutcome` too.

csp/handlers/processes.go
```diff
@@ -196,68 +202,285 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+func (c *CSPHandlers) authorizeQuestion(
+	sc *signContext, electionID internal.HexBytes,
+) (internal.HexBytes, *errors.Error) {
+	// an empty id cannot name an election, and db.QuestionByUpstreamID reports it as
+	// ErrInvalidData, which the branch below would surface to the client as a 500. The message is
+	// field-neutral because the single-sign endpoint sends this same id as "electionId".
+	if len(electionID) == 0 {
+		return nil, asErr(errors.ErrMalformedBody.With("missing election id"))
+	}
+	question, err := c.mainDB.QuestionByUpstreamID(electionID)
```
elboletaire, line 335: Per-ballot `QuestionByUpstreamID` means up to `db.MaxQuestionsPerProcess` (100) **sequential** Mongo round trips per request on a public endpoint. The handler already knows every ballot must belong to this process: one `c.mainDB.QuestionsByProcess(oid)` call (already used by `ProcessCheckHandler` and `ProcessSignInfoHandler` in this file) fetches all questions in a single query; resolve each ballot against a `map[string]*db.VotingProcessQuestion` keyed on `string(UpstreamID)`. Error semantics are preserved: not-found and wrong-process already collapse into the same 401 "election not found in process", the empty-id 400 runs before any DB access, and `memberEligibleForQuestion` takes the question struct unchanged. The single-sign path can keep its one-query lookup.

csp/handlers/processes.go
```diff
@@ -196,68 +202,285 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+type signContext struct {
+	process  *db.VotingProcess
+	memberID string
+	weight   internal.HexBytes
+}
+
+// resolveSignContext runs the per-request half of a ballot signing request: load the voting
+// process, check the auth token is verified and anchored to it, resolve the org member behind
+// it and compute their census weight. It writes the proper error and returns false on failure.
+func (c *CSPHandlers) resolveSignContext(
```
elboletaire, line 266: Heads-up on a client-visible precedence change from the refactor: the old inline `ProcessSignHandler` checked the question (`election not found in process` / eligibility, both 401) **before** resolving the member, census, and weight; `resolveSignContext` now runs member/census/zero-weight first. A request that is wrong on both axes (bad electionId AND deleted member / missing census / zero-weight voter) now returns 404 `ErrUserNotFound` / `ErrCensusNotFound` / `ErrZeroWeightVoter` where it previously returned the 401. No in-repo test depends on the old order, so probably fine — flagging so the change is deliberate rather than accidental.

csp/handlers/processes.go
```diff
@@ -196,68 +202,285 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+		seen[string(item.UpstreamID)] = i
+		upstreamID, sErr := c.authorizeQuestion(sc, item.UpstreamID)
+		if sErr != nil {
+			sErr.Withf("at index %d", i).Write(w)
+			return
+		}
+		// the address is signed into the CA bundle and pinned as the election's consumer, after
+		// which a different address is rejected forever — so validate its length here. HexBytes
+		// accepts any length and pads odd input, which is exactly the silent corruption to stop.
+		if len(item.Address) != common.AddressLength {
```
elboletaire: The 20-byte rule (and its "pinned as the election's consumer" rationale comment) now lives twice: here and in `parseAddress` (handlers.go:344), both added by this PR. This is exactly the invariant that must not drift — a future change to one copy lets a malformed address through the other path and permanently locks the voter out. Suggest a single `validateVoterAddress(internal.HexBytes) *errors.Error` that `parseAddress` and this loop both call.

csp/handlers/processes.go
```diff
@@ -4,20 +4,26 @@ import (
...
 	"github.com/vocdoni/saas-backend/internal"
 	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.vocdoni.io/dvote/log"
 	"go.vocdoni.io/dvote/vochain/state"
 )
 
+// apiErr lifts an API error to a pointer, so a helper can signal "no error" with nil. Named
+// asErr (not apiErr) because handlers.go has several `apiErr, ok := err.(errors.Error)` locals
+// that would shadow a same-named helper.
+func asErr(e errors.Error) *errors.Error { return &e }
```
elboletaire: The godoc starts with the wrong name — "apiErr lifts..." for a function named `asErr` (the next sentence even explains it was deliberately *not* named `apiErr`). Should read "asErr lifts...". Also, this is the third spelling of the same lift-to-pointer idiom (`asPtr` in api/process_vote.go:486, and the inline temporaries in the moved `apicommon.DecodeCappedJSON`); a single exported `errors.Ptr(e Error) *Error` next to the type would serve all three.

elboletaire (changes requested):
Fable 5 here, controlled by elboletaire. Reviewed the sign-batch endpoint end to end (8 finder angles + per-finding verification). No blocker inside the new authorization logic itself — the all-or-nothing gate, dedupe, and body caps hold up. Seven findings below as inline comments: the two worth acting on before merge are the `sign_failed` catch-all that tells a client to retry an auth failure forever, and a pre-existing signer-lock leak in `csp/sign.go` that the new `already_signing` retry contract turns into a permanent lockout.

csp/handlers/processes.go
lucasmenendez: Fixed in 41ccfe5, at the root rather than by unlocking on the error path: the `Verified` check now runs **before** the lock is taken (the token data is already loaded two lines up), which deletes the reachable leak entirely. The error paths that remain after locking (`proto.Marshal`, salt size) now return `authTokenData.UserID` alongside the error, so `Sign`'s deferred unlock always hashes the key that was actually locked — belt and braces.

Regression test in `csp/sign_test.go`: a sign with an unverified token fails, asserts the pair is not locked, then verifies the token and asserts the retry succeeds instead of reporting `ErrUserAlreadySigning` forever.

You were right that the new retry contract upgraded this from latent to user-facing — thanks for connecting those two.

csp/handlers/processes.go
lucasmenendez: Fixed in 41ccfe5 with the distinct code: `signOutcome` maps `ErrInvalidAuthToken`/`ErrAuthTokenNotVerified` to **`auth_invalid`** ("the auth token is no longer valid; authenticate again"), and the signing loop **short-circuits** on it — the remaining entries are stamped with the same outcome instead of each burning a doomed `Sign` round trip. Kept the 200 rather than aborting with 401: by that point earlier ballots may have consumed real signing slots, and the per-item response is what tells the client which ones. The codes are now named constants and the godoc spells out retryable vs not.

csp/handlers/processes.go
lucasmenendez: Fixed in 41ccfe5 — `signAndRespond` now routes through `signOutcome`: known authorization outcomes stay 401 with the sanitized message, the catch-all becomes a 500, and the raw joined error goes to the log. Since `BundleSignHandler` shares the helper, the deprecated path gets the same treatment for free.

This surfaced a latent misclassification that `TestCSPVoting` caught immediately: reusing a token with a *different* address was coming back as `ErrSign` (→ 500 under the new mapping), but it is an authorization rejection — the slot is pinned to the first address. `finishSaltedKeySigner` now maps that `db.ErrInvalidData` to `ErrProcessAlreadyConsumed`, so it stays a 401/`already_consumed` everywhere.

csp/handlers/processes.go
lucasmenendez: Fixed in 41ccfe5 as suggested: one `QuestionsByProcess(oid)` load, ballots resolved against a `map[string]*db.VotingProcessQuestion`. `authorizeQuestion` split into the DB lookup half (kept by single-sign) and `authorizeLoadedQuestion` (shared), so the authorization logic itself still has one home. Error semantics preserved: empty-id 400 before the map, map-miss folds into the same 401.

csp/handlers/processes.go
lucasmenendez: Deliberate, confirming on the record. Resolving member/census/weight once per request is the reason the batch endpoint exists, and the single-sign path shares that shape so the two stay one code path. The orderings differ only for requests that are wrong on two axes at once, both replies are rejections, and no client or test depends on which rejection wins. Not worth a second per-ballot pass to restore the old precedence.

csp/handlers/processes.go
lucasmenendez: Fixed in 41ccfe5: `validateVoterAddress` in `csp/handlers`, called by both `parseAddress` and the batch validation pass, with the pinned-consumer rationale on the helper — agreed this is exactly the invariant that must not fork.

csp/handlers/processes.go
lucasmenendez: Fixed in 41ccfe5, taking the full suggestion: `errors.Ptr(e Error) *Error` now lives next to the type, and all three spellings are gone — `asErr` (whose godoc I had indeed left saying "apiErr"), `asPtr` and its nine call sites, and the temporaries in `apicommon.DecodeCappedJSON`.

csp/handlers/processes.go
```diff
@@ -228,85 +226,355 @@ func (c *CSPHandlers) ProcessAuthResendHandler(w http.ResponseWriter, r *http.Re
...
+	}
+	byUpstream := make(map[string]*db.VotingProcessQuestion, len(questions))
+	for i := range questions {
+		byUpstream[string(questions[i].UpstreamID)] = &questions[i]
+	}
+	ballots := make([]authorizedBallot, len(req.Ballots))
+	seen := make(map[string]int, len(req.Ballots))
+	for i, item := range req.Ballots {
+		// a repeated election would contend with itself on the per-(user, election) signer
+		// lock, and a voter has at most one ballot per question anyway.
```
elboletaire, line 464: **A transient Mongo failure on any ballot aborts the rest of the batch and forces the voter to re-authenticate.** This is the one blocking item, and it's a regression introduced by the fix to my earlier comment on `signOutcome`.

The full chain, each hop read rather than assumed:

1. `db/csp.go:281-285` — `fetchCSPAuthFromDB` maps only `mongo.ErrNoDocuments` to a typed error; every other driver error is returned raw.
2. `db/csp.go:79` — `CSPAuth` wraps the call in a `defaultTimeout` context, so a slow or unreachable Mongo produces `context.DeadlineExceeded` down that same raw path.
3. `csp/sign.go:58-61` — `prepareSaltedKeySigner` collapses *any* error from `CSPAuth` into `ErrInvalidAuthToken`.
4. `csp/handlers/processes.go:499` — `signOutcome` maps `ErrInvalidAuthToken` to `signCodeAuthInvalid`.
5. This branch — stamps every remaining ballot with "the auth token is no longer valid; authenticate again" and `break`s.

So: a 10-question batch, Mongo hiccups on ballot 1, and the voter is pushed through a fresh OTP round trip (email/SMS cost, plus the resend cooldown) while holding a perfectly valid token — with nine ballots never even attempted.

The comment above this branch states the premise precisely — *"the token was valid when `resolveSignContext` ran, so it was deleted or unverified mid-batch"* — and that premise is sound. It just isn't what the code tests, because step 3 makes "storage was briefly unavailable" indistinguishable from "token is gone".

Before 41ccfe5 this case landed in `sign_failed`: per-ballot, retryable, siblings unaffected. That was the right outcome for it, and it's what the fix took away. Two ways out, either is fine:

- Narrow the short-circuit to `signCodeAuthInvalid` raised by `ErrAuthTokenNotVerified` only — that one genuinely does doom every remaining ballot.
- Or fix it at the root in `prepareSaltedKeySigner` (see my other comment) so a storage failure never becomes an auth verdict in the first place. I'd prefer this one: `finishSaltedKeySigner` has the identical collapse at `csp/sign.go:112-115`, so the root fix covers both.

I traced this rather than reproduced it live — `csp.CSP.Storage` is a concrete `*db.MongoStorage` rather than an interface, so there's no cheap stub and a real repro means killing the container mid-request. Flagging that so you can weigh it, though I don't think the chain has anywhere to hide.

csp/sign.go
```diff
@@ -59,6 +59,13 @@ func (c *CSP) prepareSaltedKeySigner(token, address, processID, weight internal.
 	if err != nil {
 		return nil, nil, nil, ErrInvalidAuthToken
```
elboletaire: This is the root of the batch-abort problem I flagged in `csp/handlers/processes.go`. Every failure mode of `c.Storage.CSPAuth(token)` becomes `ErrInvalidAuthToken` here — token genuinely absent, but equally a Mongo timeout, a dropped connection, or a decode error, since `fetchCSPAuthFromDB` (`db/csp.go:281-285`) only special-cases `mongo.ErrNoDocuments` and returns everything else untouched.

That conflation was harmless while every `csp.Sign` failure produced the same 401 on a single-ballot endpoint. It stopped being harmless once `auth_invalid` acquired cross-ballot meaning in the batch handler: "storage blinked" now reads as "this voter's session is dead" and takes the remaining ballots with it.

Suggested shape — distinguish the two at the point where the information still exists:

```go
authTokenData, err := c.Storage.CSPAuth(token)
if err != nil {
	if errors.Is(err, db.ErrTokenNotFound) {
		return nil, nil, nil, ErrInvalidAuthToken
	}
	return nil, nil, nil, errors.Join(ErrSign, err)
}
```

`ErrSign` maps to `sign_failed` through `signOutcome`'s default branch, which is the retryable per-ballot outcome this case wants, and it keeps the raw error available to the log without widening the public body.

`finishSaltedKeySigner` has the same collapse at lines 112-115 and would want the same treatment — worth doing together so the two paths don't drift, which is the argument that produced `validateVoterAddress` earlier in this review.

elboletaire (changes requested):

Everything from the previous rounds checks out as fixed — I re-read all eleven threads against the code, not just the replies: the `ErrInvalidData`→400 mapping, the `weightBytes`/`SetUint64` overflow fix, the `logSignFailure` noise split, the `UpstreamID` rename, the routes.go wording, the lock leak (`Verified` moved above `c.lock`), the `signAndRespond` 401/500 split, the N+1 collapse into `QuestionsByProcess` + `byUpstream`, the shared `validateVoterAddress`, and `errors.Ptr` with `asErr` gone. Build and vet are clean; `go test ./csp/...` passes, and `TestProcessCSP` + `TestProcessCSPSignBatch` both pass against real Voconed + Mongo containers.

One thing holds this up, and it comes out of the fix to my own comment on `signOutcome`. Asking for `ErrInvalidAuthToken`/`ErrAuthTokenNotVerified` to stop hiding in `sign_failed` was right, but 41ccfe5 routed them into `auth_invalid`, which *also* aborts the rest of the batch — and transient storage failures reach that same code. For that case the behaviour regressed: it used to be one retryable per-ballot `sign_failed`, and now it kills every remaining ballot and pushes the voter into a fresh OTP flow. Details inline.

Nothing else here is blocking. Two follow-ups I'd take in this PR or the next one, your call: `db.ErrInvalidData` from `ConsumeCSPProcess` surfacing as `already_consumed` tells clients "not retryable" when the correct action is the opposite (re-send with the address `sign-info` reports for that question), and the `SignBatchResult` doc comment still lists only three codes — `auth_invalid` is missing, though swagger has all four. The PR body is also now inaccurate on two points worth a line each: the single-sign path *did* change status codes (401→500 for the catch-all, which I asked for), and the deprecated `POST /process/bundle/{bundleId}/sign` inherits the new hard 20-byte address check through `parseAddress`.

csp/sign.go
lucasmenendez: Fixed at the root in 1054fd0, taking your preferred option — and your exact shape. Both `prepareSaltedKeySigner` and `finishSaltedKeySigner` now split on `db.ErrTokenNotFound`:

```go
if errors.Is(err, db.ErrTokenNotFound) {
    return ..., ErrInvalidAuthToken
}
return ..., errors.Join(ErrSign, err)
```

Joined `ErrSign` flows through `signOutcome`'s default branch → `sign_failed`, per-ballot and retryable, with the raw error in the log. Done in both helpers in the same commit, per the validateVoterAddress precedent.

`TestSignOutcome` now pins the joined shape explicitly: `errors.Join(csp.ErrSign, <raw>)` → `sign_failed`, with a comment naming why (a transient Mongo error must not abort the batch). The live-fault path stays trace-verified as you noted — `CSP.Storage` being a concrete type makes a cheap stub impossible; the sentinel-level split is now the tested contract.

csp/handlers/processes.go
lucasmenendez: Fair catch — this was a regression from my own round-2 fix, and your trace was exact at every hop. Fixed in 1054fd0 via the root option (details on your `csp/sign.go` thread): a storage failure can no longer become `ErrInvalidAuthToken`, so `auth_invalid` — and with it the short-circuit — now fires only when the token is genuinely gone (`db.ErrTokenNotFound`) or unverified mid-batch. The premise the branch comment states is now actually what the code tests. A Mongo blip on ballot 1 is back to a per-ballot retryable `sign_failed` with the nine siblings still attempted.

Also took both non-blocking follow-ups in the same commit:
- **`address_mismatch`**: new `csp.ErrAddressMismatch` sentinel for `ConsumeCSPProcess`'s pinned-address rejection, mapped to its own code with the fix in the message ("re-send using the address reported by sign-info") — you were right that `already_consumed` told clients the exact opposite of the truth. Integration test asserts the code on a real re-sign with a different address.
- **Doc drift**: `SignBatchResult`'s comment now points at the `signCode*` constants as the source of truth instead of hand-listing a stale subset, and the endpoint godoc enumerates all five codes.

PR body corrected on both points you flagged (single-sign 401→500 for the catch-all; the bundle endpoint inheriting the 20-byte address check).

elboletaire (approved):

Approved. 1054fd0 fixes the blocker at the root, and I re-derived the result rather than reading the commit message: a storage failure can no longer reach an auth verdict, because `prepareSaltedKeySigner` and `finishSaltedKeySigner` both split on `db.ErrTokenNotFound` and join everything else into `ErrSign`, which lands in `signOutcome`'s default branch as the retryable per-ballot `sign_failed`. The batch short-circuit is untouched and now fires only on the case its own comment describes — token genuinely gone, or unverified mid-batch. Doing both helpers in one commit is the right call; leaving `finishSaltedKeySigner` behind is exactly the drift `validateVoterAddress` was created to prevent.

Checked for regressions around the new joined error, since widening what `ErrSign` carries is the kind of change that leaks sideways:

- **Nothing else switches on these sentinels in the sign path.** The two remaining direct-equality switches on `csp.ErrInvalidAuthToken` (`csp/handlers/handlers.go:278`, `csp/handlers/processes.go:767`) are both on the `ResendChallenge` path, so a joined error can't reach a `switch err` and silently fall through to a wrong branch.
- **`signOutcome` ordering is safe.** `ErrAddressMismatch` is a fresh sentinel that wraps nothing, and `errors.Join(ErrSign, err)` can only carry what `CSPAuth` returns — driver errors, never one of the `csp` sentinels — so no joined error can match an earlier case by accident.
- **The lock fix still holds.** The new early return sits above `c.lock`, so it returns a nil userID exactly like the `ErrInvalidAuthToken` return it replaced; `Sign`'s deferred `unlock` deletes a key that was never stored, which is a no-op. Every return below the lock still carries `authTokenData.UserID`.
- **No raw detail escapes.** Both the single-sign and batch paths render `signOutcome`'s sanitized message; the joined driver error only reaches the log.
- **`address_mismatch` keeps its status code.** It is not `signCodeFailed`, so `signAndRespond` still answers 401 on the single-sign path — same code as the `already_consumed` it replaces, with a message that now names the recovery.
- The three storage touchpoints in the sign path are finally consistent: `IsCSPProcessConsumed` already routed its storage failures to `ErrSign`, and `CSPAuth` now does the same in both helpers.

The two non-blocking follow-ups landed too, better than I asked. `address_mismatch` as its own sentinel and code is the right shape — I had suggested only fixing the mislabel, and a distinct code with the recovery in the message is more useful to clients. Pointing `SignBatchResult`'s comment at the `signCode*` constants instead of hand-listing them removes the drift that caused the original gap. Test coverage matches: `TestSignOutcome` pins the joined shape explicitly, `TestFinishSaltedKeySigner` covers the pinned-address rejection, and `TestProcessCSPSignBatch` asserts `address_mismatch` end to end with an empty signature.

Both PR-body notes from the last round are now in the description, including the 500-for-internal-failures change. One consequence worth being aware of rather than acting on: this commit slightly widens that 500: a storage failure on `POST /processes/{processId}/sign` used to answer 401 and now answers 500. That is the correct reading of a Mongo blip and it follows from the fix I asked for, so no change requested — just the sort of thing worth knowing before it shows up in an alerting threshold.

Build, vet and `scripts/check-qt-patterns.sh` are clean locally, `./csp/...` passes, and both lint checks plus swagger generation are green on CI. The `test` job was still running when I approved, so that one is on CI to confirm.

Issue #628, opened 2026-07-30
members: report the async bulk import's census propagation results


lucasmenendez:
Follow-up promised in the #621 review thread (LOW 4).

The sync path of `POST /organizations/{orgAddress}/members` surfaces `propagateMembersToCensuses` results to the caller (`censusJobIds` + `errors`). The async path cannot: by the time propagation runs, the job's `CompleteJob` has already fired, and `CompleteJob` `$set`s `errors` wholesale — a second call would clobber the import errors rather than merge the propagation ones. So `api/org_members.go` (the goroutine after the progress loop) deliberately discards the propagation result; a member imported into a live census whose resize could not be enqueued is only visible in the server log.

Fix sketch, from the review discussion:

- reorder so `CompleteJob` runs **after** propagation and merges its errors — this also fixes the currently misleading `completedAt` (stamped before the propagation work is done);
- alternatively (or additionally) a `db.Job` schema field for propagation results, with merge rather than `$set` semantics in `CompleteJob`;
- the job's response shape gains the same `censusJobIds`/`errors` the sync path already answers.

Refs #621.

PR #635, opened 2026-08-05
feat(db): index the census revocation cascade (cspTokens.userid, votingProcesses.censusId)

### LINKED ISSUE #629: db: index the census revocation cascade (cspTokens.userid, votingProcesses.censusId)
Follow-up promised in the #621 review thread (INFO 4), deferred there in commit `aa74686` because migration number 0018 was contended by three in-flight branches — not because the cost went unnoticed.

Since #621, every memberbase mutation that reaches the revocation or propagation cascade runs:

- `votingProcesses.Find({censusId: {$in: ...}})` (`db/census_revocation.go`, `VotingProcessesByCensus`) — a full collection scan; 0017's index leads with `orgAddress` and cannot serve a `censusId` predicate;
- `cspTokens.DeleteMany({userid: {$in: ...}})` (`revokeWrites` step 3) — `cspTokens` has no index beyond `_id`, so this scans too (as do the pre-existing `LastCSPAuth`/`DeleteCSPAuthByBundle`).

Neither is a correctness problem, and the collections are small today. Once migration numbering settles, add a fresh migration with:

- `votingProcesses`: `{censusId: 1}`
- `cspTokens`: `{userid: 1}`

The third index the original 0018 carried (`processesQuestions.eligibleMemberIds`) was redundant with 0017's `processId` index and should stay dropped.

Refs #621.

lucasmenendez:
Closes #629. Refs #621.

**Stacked on #621** (`feat/memberbase-census-sync`) — merge that first; this PR then carries only the one migration commit.

Migration 0019 restores the indexes from the migration #621 dropped in `aa74686` while slot 0018 was contended (0018 is now #615's unique question-slot index). Following review, the keys are chosen to cover **every** `cspTokens` access path rather than only the cascade's own.

### `cspTokens` — two indexes, previously nothing beyond `_id`

| Index | Serves |
|---|---|
| `{userid: 1, bundleid: 1, createdat: -1}` | `LastCSPAuth` (the hot voter-auth path) in full, sort included — and the revocation cascade's `DeleteMany({userid: $in})` on its `userid` prefix |
| `{bundleid: 1}` | `DeleteCSPAuthByBundle` and the two `Count*ByBundle` helpers |

The compound key means one index serves both the cascade and the voter-auth read. `bundleid` needs its own: those three queries predicate on `bundleid` with no `userid`, and a non-leading field is not a usable prefix. `DeleteCSPAuthByBundle` runs **once per bundle in a loop** on org teardown (`api/organizations_managed.go`) and GDPR erasure (`db/user_erase.go`), so those are the scans that repeat.

### `votingProcesses` — `{censusId: 1}`

0017's index leads with `orgAddress` and cannot serve `VotingProcessesByCensus`' `censusId` predicate. Harmless before #621 lands, load-bearing after.

The third index the original migration carried (`processesQuestions.eligibleMemberIds`) was redundant with 0017's `processId` index and stays dropped, per the issue.

### Rollout cost

The build runs **inline during `RunMigrationsUp`** on the startup path, under a **single 10-minute context shared by all pending migrations** (`db/migrations.go`) — so on a database jumping from <18 it shares that budget with 0018's `dedupeQuestionSlots` data pass. On MongoDB ≥ 4.2 index builds are hybrid, so writes are not blocked and the risk is the shared timeout on a large `cspTokens`, not availability. Worth knowing before deploying against a big collection; no code change implied.

### Reversibility

Index-only, so genuinely reversible unlike the data-bearing collections. The down drops all three and tolerates both `IndexNotFound` (27) and `NamespaceNotFound` (26) — the driver forgives 26 only on `IndexView.List`, so `DropOne` needs it spelled out — inspecting the error with `errors.As` per CLAUDE.md.

`TestCensusRevocationIndexesMigration` asserts the index **key specs** rather than their auto-derived names (the compound field order and the descending `createdat` are what make it serve both paths), re-runs `Up` over the already-migrated state to pin that a replay is a no-op, and asserts all three are gone after `Down`.

## Verification

```
golangci-lint run                # 0 issues
./scripts/check-qt-patterns.sh   # clean
go test -count=1 ./db/           # 251 passed
```

The test was also run against the previous single-`{userid:1}` shape and fails there. No API change, so no swagger regeneration.

### DIFF
--- a/db/migration_census_revocation_indexes_test.go
+++ b/db/migration_census_revocation_indexes_test.go
@@ -1,0 +1,68 @@
+package db
+
+import (
+	"context"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+	"github.com/vocdoni/saas-backend/migrations"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+// indexKeys maps a collection's index names to their key specs. The key spec is what the queries
+// actually depend on — a name is derived from it, so asserting only the name would pass on an
+// index whose field order or direction had drifted.
+func indexKeys(c *qt.C, coll *mongo.Collection) map[string]bson.D {
+	cursor, err := coll.Indexes().List(context.Background())
+	c.Assert(err, qt.IsNil)
+	var indexes []struct {
+		Name string `bson:"name"`
+		Key  bson.D `bson:"key"`
+	}
+	c.Assert(cursor.All(context.Background(), &indexes), qt.IsNil)
+	keys := make(map[string]bson.D, len(indexes))
+	for _, idx := range indexes {
+		keys[idx.Name] = idx.Key
+	}
+	return keys
+}
+
+// TestCensusRevocationIndexesMigration asserts migration 0019 gives every cspTokens access path an
+// index and the census-to-processes lookup its own, that re-running it is a no-op, and that its
+// down migration drops all three.
+//
+// The key specs are asserted rather than the names: the compound index's field order and its
+// descending createdat are what let it serve LastCSPAuth's sort and still prefix-serve the
+// revocation cascade, and a name alone would not pin either.
+func TestCensusRevocationIndexesMigration(t *testing.T) {
+	c := qt.New(t)
+	ctx := context.Background()
+	mig, ok := migrations.AsMap()[19]
+	c.Assert(ok, qt.IsTrue)
+	database := testDB.DBClient.Database(testDB.database)
+	c.Cleanup(func() {
+		// restore the migrated state for the rest of the suite
+		c.Assert(mig.Up(ctx, database), qt.IsNil)
+	})
+
+	// the test database is migrated on init
+	cspAuth := bson.D{{Key: "userid", Value: int32(1)}, {Key: "bundleid", Value: int32(1)}, {Key: "createdat", Value: int32(-1)}}
+	c.Assert(indexKeys(c, testDB.cspTokens)["userid_1_bundleid_1_createdat_-1"], qt.DeepEquals, cspAuth)
+	c.Assert(indexKeys(c, testDB.cspTokens)["bundleid_1"], qt.DeepEquals, bson.D{{Key: "bundleid", Value: int32(1)}})
+	c.Assert(indexKeys(c, testDB.votingProcesses)["censusId_1"], qt.DeepEquals, bson.D{{Key: "censusId", Value: int32(1)}})
+
+	// re-running over the already-migrated state changes nothing: an identical spec and name is a
+	// server-side no-op, so a redeploy that replays the migration is safe
+	c.Assert(mig.Up(ctx, database), qt.IsNil)
+	c.Assert(indexKeys(c, testDB.cspTokens)["userid_1_bundleid_1_createdat_-1"], qt.DeepEquals, cspAuth)
+
+	c.Assert(mig.Down(ctx, database), qt.IsNil)
+	after := indexKeys(c, testDB.cspTokens)
+	_, ok = after["userid_1_bundleid_1_createdat_-1"]
+	c.Assert(ok, qt.IsFalse)
+	_, ok = after["bundleid_1"]
+	c.Assert(ok, qt.IsFalse)
+	_, ok = indexKeys(c, testDB.votingProcesses)["censusId_1"]
+	c.Assert(ok, qt.IsFalse)
+}
--- a/migrations/0019_census_revocation_indexes.go
+++ b/migrations/0019_census_revocation_indexes.go
@@ -1,0 +1,72 @@
+package migrations
+
+import (
+	"context"
+	stderrors "errors"
+	"fmt"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+func init() {
+	AddMigration(19, "census_revocation_indexes", upCensusRevocationIndexes, downCensusRevocationIndexes)
+}
+
+// upCensusRevocationIndexes adds the indexes behind the census revocation cascade (#621), promised
+// as a follow-up in #629. Each of these lookups was a collection scan before: cspTokens carried no
+// index beyond _id, and 0017's votingProcesses index leads with orgAddress.
+func upCensusRevocationIndexes(ctx context.Context, database *mongo.Database) error {
+	// Two indexes cover every way cspTokens is queried.
+	//
+	// The compound one is keyed for LastCSPAuth — filter {userid, bundleid}, sort {createdat: -1},
+	// the hot voter-auth path — and its userid prefix serves the revocation cascade's
+	// DeleteMany({userid: $in}) just as a bare {userid: 1} would.
+	//
+	// bundleid alone is the second: DeleteCSPAuthByBundle and the two Count*ByBundle helpers
+	// predicate on bundleid with no userid, so the compound index cannot serve them (a non-leading
+	// field is not a usable prefix). The delete runs once per bundle in a loop on org teardown and
+	// GDPR erasure, so those are the scans that repeat.
+	if _, err := database.Collection("cspTokens").Indexes().CreateMany(ctx, []mongo.IndexModel{
+		{Keys: bson.D{
+			{Key: "userid", Value: 1},
+			{Key: "bundleid", Value: 1},
+			{Key: "createdat", Value: -1},
+		}},
+		{Keys: bson.D{{Key: "bundleid", Value: 1}}},
+	}); err != nil {
+		return fmt.Errorf("failed to create indexes on cspTokens: %w", err)
+	}
+	// the access path from a census back to the processes built on it (VotingProcessesByCensus,
+	// arriving with #621); 0017's index leads with orgAddress and cannot serve a censusId predicate.
+	if _, err := database.Collection("votingProcesses").Indexes().CreateMany(ctx, []mongo.IndexModel{
+		{Keys: bson.D{{Key: "censusId", Value: 1}}},
+	}); err != nil {
+		return fmt.Errorf("failed to create indexes on votingProcesses: %w", err)
+	}
+	return nil
+}
+
+// downCensusRevocationIndexes drops the indexes again. They carry no data, so unlike the
+// data-bearing collections this migration is genuinely reversible.
+func downCensusRevocationIndexes(ctx context.Context, database *mongo.Database) error {
+	for _, idx := range []struct {
+		collection string
+		name       string
+	}{
+		{"cspTokens", "userid_1_bundleid_1_createdat_-1"},
+		{"cspTokens", "bundleid_1"},
+		{"votingProcesses", "censusId_1"},
+	} {
+		if _, err := database.Collection(idx.collection).Indexes().DropOne(ctx, idx.name); err != nil {
+			// Nothing to drop is not a failure to drop: IndexNotFound (27) means the up never got
+			// far enough to create it, NamespaceNotFound (26) that the collection itself is gone.
+			// The driver only forgives 26 on IndexView.List, so DropOne needs it spelled out here.
+			var cmdErr mongo.CommandError
+			if !stderrors.As(err, &cmdErr) || (cmdErr.Code != 27 && cmdErr.Code != 26) {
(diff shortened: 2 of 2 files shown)


elboletaire (changes requested):
openai here, controlled by elboletaire.

Requesting changes. These are the final confirmed issues after cross-review between `deepseek/deepseek-v4-pro` and `openai-codex/gpt-5.6-sol`.

### High

- **`api/organization_groups.go`**
  Group update propagation uses raw `AddMembers` instead of only actually-new group members, so an idempotent duplicate add can silently undo a prior manual census revocation.
  **Recommendation:** propagate only a deduplicated `actuallyAdded` set derived from members newly entering the group.

- **`api/census_propagation.go`, `api/org_members.go`**
  `autoGroupCensuses` swallows operational DB errors as “no censuses,” so member creation can skip preflight and propagation and still report success.
  **Recommendation:** return an error for real lookup failures and abort before member writes, or persist propagation for retry.

### Medium

- **`db/census_revocation.go`**
  Revocation writes commit before resize-target discovery (`emptiedQuestions`). If that later read fails, the published questions needing resize can be lost, and retry cannot reconstruct them from the already-pruned state.
  **Recommendation:** preserve resize work durably before returning after committed revocation; for example by precomputing/persisting targets or using an outbox/transactional pattern.

- **`api/org_members.go`**
  Async member import marks the job complete before census propagation, and drops propagation errors/job IDs, so imports can appear successful while live censuses remain inconsistent.
  **Recommendation:** include propagation in job completion semantics and persist/report propagation errors and spawned resize jobs.

### Low

- **`errors/errors_definition.go`, `api/census_guard.go`**
  Internal/public naming still says “already voted” / `votedMemberIds` even though the semantics are “already signed for,” which is misleading for API consumers and UIs.
  **Recommendation:** rename toward “signed” semantics, or add a compatibility/deprecation path if the wire format must remain stable.

- **`db/census_revocation.go`**
  The `emptiedQuestions` Mongo filter depends on null/missing-field equivalence, but that behavior is only encoded in tests, not documented near the query itself.
  **Recommendation:** add an explicit code comment explaining that the query intentionally matches null, empty array, and missing field encodings.

lucasmenendez:
Thanks — all six findings landed on this PR, but none of them are in its diff: #635 is stacked on #621 and contains only `migrations/0019_census_revocation_indexes.go` plus its test. Every file named in the review (`api/organization_groups.go`, `api/census_propagation.go`, `db/census_revocation.go`, `errors/errors_definition.go`, `api/org_members.go`) belongs to the base branch, so I have fixed them **on #621** and left this PR untouched. Full detail is in [my reply there](https://github.com/vocdoni/saas-backend/pull/621#issuecomment-5202068495); the summary:

**Acted on: H1, H2, M1 (which subsumes L2), L1. Answered: M2.** Each was verified against the code rather than taken as read, and every fix has a test that fails on the commit before it.

| Finding | Verdict | Commit |
|---|---|---|
| **H1** group re-add undoes a revocation | confirmed | `5481b3c` |
| **H2** `autoGroupCensuses` swallows lookup errors | confirmed | `0bd78f3` |
| **M1** resize targets lost after commit | confirmed | `499738e` |
| **L2** undocumented null/empty filter | — | deleted by M1's fix |
| **L1** "voted" naming | confirmed | `24fc88c` |
| **M2** async import ordering | deliberate, tracked | #628 |

Two were worse than the write-up suggested. **H2**'s swallowed error does not merely skip the preflight — it creates a member who is silently not a voter, leaving the memberbase and the censuses permanently out of step with no job to poll and nothing to retry from, behind a 200. And **M1**'s post-commit read carries a comment saying a failure past that point "must not discard it or the caller's resize is lost for good", while the next line does exactly that (`return 0, nil, err`); the mitigation had been applied to the recount below it but not to the read it was written about.

**L2 needed no separate fix.** I addressed M1 by deciding emptiness *before* the write rather than hardening the failing read, which deletes the `emptiedQuestions` re-read and its `$in {nil, []}` filter outright — net −19 lines. Worth flagging: the test L2 credited with encoding that behaviour was structurally vacuous. In all three of its cases the assertion loop ran over an empty `emptied` list, so it would have passed with the filter deleted entirely. It is replaced with a positive test of the strict-subset boundary.

**One deliberate behaviour change**, in H2: a real auto-group lookup failure now fails the request retryably instead of answering 200. That trades availability for consistency on a write whose whole purpose is keeping the two in step.

**L1** is marked `!` as a wire rename, but needs no compatibility path: error 40173 and `data.votedMemberIds` were introduced by #621 and are absent from `main`, so nothing shipped consumes either name.

**M2** is deliberate rather than overlooked — `f3ae229` documents it at the call site with the `CompleteJob` clobber mechanics (it `$set`s `errors` wholesale, so a second write destroys the import errors) and points at **#628**, filed for it. The real fix needs a `db.Job` schema field and a response-shape change, so it is its own PR.

Verification on #621: build, vet, `golangci-lint` (0 issues), qt-pattern check, `go mod tidy` (no diff), `make swagger` (only the L1 rename), and `go test ./...` — 17 packages, 0 failures. This PR has been rebased on the updated base and still carries its single migration commit.

elboletaire (changes requested):

Scoped to this PR's own diff — commit `9e5a553`: `migrations/0019_census_revocation_indexes.go` and `db/migration_census_revocation_indexes_test.go`. The stack it rides on is reviewed separately in #621.

The migration itself is sound: both index specs conflict with nothing (no pre-existing `userid_1` on `cspTokens` nor `censusId_1` on `votingProcesses` — checked every `Indexes()` call site across `migrations/` and `db/`), slot 19 is unique, it is a faithful restore of the dropped 0018 minus the redundant `eligibleMemberIds` index, and the down is genuinely reversible. The test proves presence-after-up / absence-after-down and follows the qt conventions.

Findings:

1. **The rationale overclaims what `{userid: 1}` serves.** The migration comment and the PR body say it fixes the scan for `LastCSPAuth` **and** `DeleteCSPAuthByBundle`. The latter filters on `bundleid` only (`db/csp.go:473`) — as do `CountCSPAuthByBundle` and `CountCSPAuthVerifiedByBundle` — so the new index cannot serve it; the org-teardown (`api/organizations_managed.go:497`) and GDPR-erasure (`db/user_erase.go:137`) loops still scan `cspTokens` whole, once per bundle. Either drop the claim or add `{bundleid: 1}` to the same migration.
2. **A compound key would be strictly better for the same single index.** `LastCSPAuth` is the hot voter-auth path — filter `{userid, bundleid}` + sort `{createdat: -1}` — and a bare `{userid: 1}` leaves the bundleid match and the sort to run after the index scan. `{userid: 1, bundleid: 1, createdat: -1}` serves it fully and still serves the revocation cascade's `DeleteMany({userid: {$in: …}})` as a prefix. Worth deciding before the index ships.
3. **Error inspection by type assertion.** The down uses `err.(mongo.CommandError)` instead of `errors.As` (project convention) — works with the current driver's unwrapped error values, breaks silently if the error is ever wrapped. It copies 0017's style, so arguably consistent; flagging anyway.
4. **The down tolerates `IndexNotFound` (27) but not `NamespaceNotFound` (26)** — if either collection is absent when Down runs, the rollback aborts. Low probability given migration ordering, but the guard is one code short of complete.
5. **The test could be slightly stronger**: it asserts only the auto-derived index *names*, and never re-runs `Up` against an already-migrated state. Idempotency is safe in practice (identical spec+name is a Mongo no-op) but unproven; asserting the `key` specs and a second `Up` returning nil would close both.
6. **Build cost is unstated**: the index build runs inline during `RunMigrationsUp` under the single shared 10-minute context covering all pending migrations, blocking startup. On MongoDB ≥ 4.2 builds are hybrid so writes are not blocked; the real risk is the shared timeout budget on a large `cspTokens`. Worth a sentence in the PR body; no code change needed.

None of this is blocking on its own — the request-changes state is for item 1 (the migration ships a comment claiming a win it does not deliver) and for settling item 2 while the index has not shipped yet.

emmdim:
Agree with @elboletaire 's comment. Especially focusing in comments 1, 2 and 4

lucasmenendez:
Thanks — this was a sharp pass, and finding 1 was right in the way that matters most: the migration shipped a comment claiming a win it did not deliver. All six are addressed in `eae6e99` (amended, so this PR keeps its single commit). Each was verified against the code first; two of your framings needed a small correction, noted below.

## 1 + 2 — the index shape, settled together

Both confirmed, and they turned out to be one decision rather than two. I enumerated every query against `cspTokens` before choosing:

| Query | Filter | Sort |
|---|---|---|
| revocation cascade (`db/census_revocation.go`) | `{userid: $in}` | — |
| `LastCSPAuth` (`db/csp.go:97`) | `{userid, bundleid}` | `{createdat: -1}` |
| `DeleteCSPAuthByBundle` (`db/csp.go:473`) | `{bundleid}` | — |
| `CountCSPAuthByBundle` (`db/csp.go:328`) | `{bundleid}` | — |
| `CountCSPAuthVerifiedByBundle` (`db/csp.go:346`) | `{bundleid, verified}` | — |
| `fetchCSPAuthFromDB` / `VerifyCSPAuth` / `IncrementCSPAuthAttempts` | `_id` | — |

So the migration now creates **`{userid: 1, bundleid: 1, createdat: -1}`** and **`{bundleid: 1}`**. The compound key serves `LastCSPAuth` fully — sort included, which the bare `{userid: 1}` left to run after the index scan — and its `userid` prefix serves the cascade exactly as before, so your point 2 costs nothing to take.

On point 1 I went with delivering the claim rather than retracting it: `{bundleid: 1}` is the second index, so `DeleteCSPAuthByBundle` and both `Count*ByBundle` helpers are now served. That one earns its keep — the delete runs **once per bundle in a loop** on org teardown (`api/organizations_managed.go:485`) and GDPR erasure (`db/user_erase.go:126`), so it was the repeating scan, not a one-off.

One correction: both `Count*ByBundle` are `Distinct("userid", …)`, not `CountDocuments`. Doesn't change the conclusion — the predicate shape is what matters and you had that right.

## 4 — NamespaceNotFound

Confirmed and fixed: the down now tolerates **26** alongside 27. Worth recording why it is one code short rather than an oversight of a known pattern — the driver forgives 26 only inside `IndexView.List` (`index_view.go:119`), and `drop` passes it straight through, so `DropOne` needs it spelled out. No other migration in the repo handles 26 either, and `mongo_helpers.go`'s `strings.Contains(err.Error(), "IndexNotFound")` does not (a missing namespace reads "ns not found"). 0019 is now the only one that does.

## 3 — errors.As

Taken. Fair to flag even though it copies house style: CLAUDE.md mandates `errors.Is`/`errors.As` for inspection, and while 0013 and 0017 both use the bare assertion, "consistent with two existing violations" is not the same as correct. `errorlint` is not enabled, so nothing would have caught it. 0019 is now the first migration using `errors.As` — deliberately the odd one out rather than silently divergent.

For the record it was not *broken*: `IndexView.drop` returns `replaceErrors(err)`, which yields an unwrapped `CommandError` value, so the assertion worked. It was brittle by convention, which is precisely the argument for changing it before something wraps.

## 5 — test strength

Both halves taken, and the first was sharper than it looked. The test now asserts the **key specs** rather than the auto-derived names — which matters much more with the compound index than it did with the bare one, since the field order and the descending `createdat` are the whole reason it serves both paths, and a name would have passed on a drifted spec. It also re-runs `Up` over the already-migrated state to pin the replay no-op, mirroring `migration_question_slots_test.go:68-72`.

Your "never re-runs Up" was strictly false in letter (there was a call in `c.Cleanup`) but exactly right in substance: that call ran *after* `Down`, so it only ever exercised create-from-scratch. I checked, and no test in `db/*_test.go` asserted index key specs before this one.

## 6 — build cost

Documented in the PR body, no code change, as you framed it. Confirmed the mechanics: the build is inline on the startup path under **one 10-minute context shared across all pending migrations** (`db/migrations.go:21-51`), so on a database jumping from <18 it shares that budget with 0018's `dedupeQuestionSlots` data pass — which is the sharper version of the risk than either migration alone.

## Verification

```
golangci-lint run                # 0 issues
./scripts/check-qt-patterns.sh   # clean
go test -count=1 ./db/           # 251 passed
```

The strengthened test was also run against the previous `{userid: 1}` shape and fails there, so it genuinely pins the new keys rather than restating them.

emmdim (approved):

elboletaire (approved):

PR #637, opened 2026-08-05
fix(census): narrow the removal guard to the elections the member would actually leave

### LINKED ISSUE #630: census: narrow the removal guard to the elections the member would actually leave
Follow-up named in #621's description under *Not fixed, deliberately*.

The removal guard (`refuseBlockedVoters`, `api/census_guard.go`) refuses deleting or removing a member when the CSP has signed for them on **any** ongoing question of the affected censuses — not only the election the request would actually strip them from. That is the safe direction, but it over-refuses: a member signed for question A cannot be removed even by an operation that only affects question B's census membership.

#621's own question-eligibility endpoint already shows the narrower shape: `refuseVotersLosingEligibility` is keyed off the edited question's own election, which is why its 409 carries no over-breadth caveat. Extending that precision to the member/group removal paths is a behavioural change (some 409s become 200s) and deserves its own PR with its own tests.

Refs #621.

lucasmenendez:
Closes #630. Refs #621.

**Stacked on #635** (`index_census_revocation`, itself stacked on #621's `feat/memberbase-census-sync`) — merge those first; this PR then carries only the one guard commit.

`blockedVoters` refused a removal when the CSP had signed for a member on **any** ongoing question of the affected censuses — the full censusIDs × memberIDs cross product. But the write it guards is pair-scoped: `revokeWrites` deletes `(census, member)` participant rows, so a member with no row in a census loses nothing there. A stale consumed `cspTokensStatus` row (they survive revocation by design — see step 4 of `revokeWrites`) for a census the member had already left could block, e.g., a member-delete whose census union only contained that census through *another* target member.

The guard now blocks a member only when they still **participate** in a census whose ongoing election they were signed for — the elections the removal would actually strip them from, the same per-election precision `refuseVotersLosingEligibility` already has.

- Checked per census, reusing the existing `CensusParticipantsByMemberIDs` / `OngoingQuestionsByCensuses` / `MembersWithUsedCSPProcesses` — no db-layer changes. The participant filter's exact `participantID` matching mirrors the revocation's own `$in` delete, so the guard refuses exactly what the write would remove.
- Question eligibility lists deliberately play **no** part in the predicate: a consumed signature is a live potential vote whatever the list says today, and the CSP sign gate checks participation, not eligibility. The `votes <= censusSize` invariant is preserved — a member with a signature but no participant row is already outside the count.
- **Behavioural change (409 → 200):** removals blocked only through censuses the member had already left now succeed. The `signedMemberIds` list (renamed from `votedMemberIds` by the base branch) is also answered once per member, in the caller's order (the old guard echoed duplicated request ids twice).
- Swag `@Description` wording tightened on the member DELETE and process-census DELETE; `make swagger` regenerated.

New tests: `TestMemberDeletionReleasedByStaleSignature` (the 409→200 flagship on the member-delete door), `TestGroupRemovalReleasedForNonParticipant` (group door), `TestBlockedVotersScopedAndDeduped` (predicate scoping, ordering and dedup, direct call). The existing refusal tests (`TestMemberDeletionRefusedForVoter`, `TestGroupCensusGuards`, `TestRemoveProcessCensusContract`, both org-scoping tests) pass unchanged and pin the still-blocks half.

## Verification

```
golangci-lint run                # 0 issues
./scripts/check-qt-patterns.sh   # clean
make swagger                     # regenerated, wording-only diff
go test -count=1 ./api/ ./db/    # 506 passed
```

### DIFF
--- a/api/census_guard.go
+++ b/api/census_guard.go
@@ -26,8 +26,10 @@ import (
 // can belong to someone who is no longer counted. The participation scoping preserves exactly
 // that: a member with a signature but no participant row is already outside the count, and only
 // the censuses that would lose a counted member can lose the invariant. Question eligibility
-// lists deliberately play no part — a consumed signature is a live potential vote whatever the
-// list says today, and the CSP sign gate checks participation, not eligibility.
+// lists deliberately play no part — eligibility is enforced once, at sign time
+// (ProcessSignHandler), and never re-checked afterwards, so a consumed signature is a live
+// potential vote whatever the list says today; participation is the only axis this removal
+// changes.
 //
 // One census is checked per iteration — the participant filter is per census, and matching it by
 // exact participantID string mirrors the $in delete in RevokeMembersFromCensuses, so the guard
@@ -71,7 +73,7 @@ func (a *API) blockedVoters(censusIDs, memberIDs []string) ([]string, error) {
 		}
 		hit, err := a.db.MembersWithUsedCSPProcesses(elections, present)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("could not resolve the members signed for on the census elections: %w", err)
 		}
 		for _, id := range hit {
 			blocked[id] = true
--- a/api/census_guard.go
+++ b/api/census_guard.go
@@ -12,31 +12,83 @@ import (
 )
 
 // blockedVoters returns the subset of memberIDs that may not be removed from the given censuses
-// because the CSP has already signed for them in a process that is still running.
+// because the CSP has already signed for them in a process the removal would actually strip them
+// from: an ongoing election of a census they still participate in. A signature on a census a
+// member has already left blocks nothing — the removal does not touch that election, so there is
+// nothing left to protect by refusing it (#630).
 //
-// "Still running" means a published question whose status is READY or PAUSED. ENDED, CANCELED and
+// "Ongoing" means a published question whose status is READY or PAUSED. ENDED, CANCELED and
 // RESULTS release their members, so once voting closes the removal goes through and erasure is
 // never permanently blocked.
 //
 // Refusing rather than silently revoking is what keeps census.Size an honest recount of live
-// participants, and votes <= censusSize true by construction: no signature the chain will accept can
-// belong to someone who is no longer counted.
+// participants, and votes <= censusSize true by construction: no signature the chain will accept
+// can belong to someone who is no longer counted. The participation scoping preserves exactly
+// that: a member with a signature but no participant row is already outside the count, and only
+// the censuses that would lose a counted member can lose the invariant. Question eligibility
+// lists deliberately play no part — a consumed signature is a live potential vote whatever the
+// list says today, and the CSP sign gate checks participation, not eligibility.
+//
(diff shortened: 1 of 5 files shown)


emmdim (approved):
### Code review

The narrowed guard is correct: the per-census participant filter mirrors the pair-scoped `$in` delete in `RevokeMembersFromCensuses` exactly, and all mutation paths that remove census membership go through `refuseBlockedVoters` (4 call sites) or `refuseVotersLosingEligibility` (eligibility narrowing). Approving. Three optional, non-blocking issues:

1. Bare `return nil, err` after `MembersWithUsedCSPProcesses`, while the two sibling checks in the same function wrap with `fmt.Errorf(...: %w)` (CLAUDE.md: "Always wrap with context").

https://github.com/vocdoni/saas-backend/blob/f4e5f21e0ba3d8134d89e4614e3aacaa7b4f7785/api/census_guard.go#L72-L75

2. The godoc says "the CSP sign gate checks participation, not eligibility", but `ProcessSignHandler` checks eligibility first (`memberEligibleForQuestion`, [csp/handlers/processes.go#L270-L274](https://github.com/vocdoni/saas-backend/blob/f4e5f21e0ba3d8134d89e4614e3aacaa7b4f7785/csp/handlers/processes.go#L270-L274)) and then participation; only the legacy `BundleSignHandler` skips eligibility. The design conclusion still holds — eligibility is enforced once at sign time and never re-checked, so participation is the right axis for the guard — but the comment misstates the reason.

https://github.com/vocdoni/saas-backend/blob/f4e5f21e0ba3d8134d89e4614e3aacaa7b4f7785/api/census_guard.go#L28-L31

3. Pre-existing (out of scope here, noting since this PR sharpens the guard's invariant claim): bundle-backed elections are invisible to the guard — `OngoingQuestionsByCensuses` only queries `votingProcesses`, so a consumed signature on a bundle election never blocks removal, even though bundle census IDs do reach `blockedVoters` via `CensusesForMembers`. Worth closing or documenting as out of scope in a follow-up.

https://github.com/vocdoni/saas-backend/blob/f4e5f21e0ba3d8134d89e4614e3aacaa7b4f7785/db/census_revocation.go#L106-L110

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

lucasmenendez:
Review addressed in 57473a7, and the branch is rebased onto the current `index_census_revocation` tip (`eae6e99`):

1. **Bare `return nil, err`** — now wrapped with context like its sibling checks: `could not resolve the members signed for on the census elections: %w`.
2. **Godoc misstating the sign gate** — corrected. `ProcessSignHandler` does check eligibility first (`memberEligibleForQuestion`) and then participation; the comment now gives the actual reason eligibility plays no part in the guard: it is enforced once at sign time and never re-checked, so a consumed signature stays a live potential vote whatever the list says today — participation is the only axis the removal changes.
3. **Bundle-backed elections invisible to the guard** — pre-existing and out of this PR's scope, filed as #639 with the close-or-document options from the review.

`go test -count=1 ./api/ ./db/` after the rebase: 508 passed; lint and qt-pattern checks clean.

PR #638, opened 2026-08-05
feat(processes)!: name the ranked and cumulative question types

### LINKED ISSUE #632: Ranked, Cumulative and Approval have no question type
Claude Opus 5 here, controlled by elboletaire.

The plan model names **six** voting types. The question API implements **two**. The other four are reachable only as a raw `ballotProtocol`, which has two present-tense consequences: the plan flags gate nothing, and a client cannot tell what kind of ballot it is holding.

Filed after #619, because #619 was a symptom of this rather than an isolated bug — see "How we got here" below.

## What exists vs what is reachable

[`db/types.go#L161-L168`](https://github.com/vocdoni/saas-backend/blob/9e811d07c7917e49ba10d39bdf477b816ef7fb78/db/types.go#L161-L168) vs [`db/const.go#L37-L38`](https://github.com/vocdoni/saas-backend/blob/9e811d07c7917e49ba10d39bdf477b816ef7fb78/db/const.go#L37-L38):

| plan flag | question type | gated by plan? | notes |
|---|---|---|---|
| `Single` | `singlechoice` | ✅ | |
| `Multiple` | `multichoice` | ✅ | |
| `Approval` | — | ❌ | probably **redundant** — the dense multichoice layout *is* approval |
| `Cumulative` | — | ❌ | budget / quadratic (`maxValue 0`, `costExponent` 1 or 2) |
| `Ranked` | — | ❌ | `uniqueValues`, `maxValue >= maxCount - 1` |
| `Weighted` | — | n/a | deliberately gated elsewhere, via `CostFromWeight` ([`subscriptions.go#L98`](https://github.com/vocdoni/saas-backend/blob/9e811d07c7917e49ba10d39bdf477b816ef7fb78/subscriptions/subscriptions.go#L98)) |

So **three flags have no type constant**, and realistically **two are genuinely missing** — the codebase already suspects `Approval` is subsumed by `Multiple`.

This is not news to the code. [`subscriptions.go#L339`](https://github.com/vocdoni/saas-backend/blob/9e811d07c7917e49ba10d39bdf477b816ef7fb78/subscriptions/subscriptions.go#L339) says it plainly:

> the remaining plan.VotingTypes flags (Ranked, Approval, Cumulative) have no entry here because db has no type constant for them, so nothing can ever resolve to one and **the flags are permissive by omission**: an org whose plan lacks Ranked can still mint a ranked ballot through a raw ballotProtocol.

and [`account/ballot.go#L102`](https://github.com/vocdoni/saas-backend/blob/9e811d07c7917e49ba10d39bdf477b816ef7fb78/account/ballot.go#L102) states the same from the other side: `ok` is false "for the shapes that have no named type — ranked, quadratic, anything using vote overwrites or weighted cost".

## Consequence 1 — plan entitlements do not gate

`OrgAllowsVotingType` can only map what has a constant. An org on a plan with `Ranked: false` mints a ranked election by sending a raw `ballotProtocol`, and nothing objects. Same for `Cumulative`. Whatever those flags were meant to sell, they currently sell nothing.

## Consequence 2 — the ballot type is not recoverable downstream

A ranked protocol and a pick-slot multichoice that fills every slot are **byte-identical**:

```
both: { maxCount: 3, maxValue: 2, uniqueValues: true, maxTotalCost: 0, costExponent: 1 }
```

They mean opposite things — for ranked the field index is the **option** and the value is its rank; for pick-slot the field index is the **slot** and the value is the chosen option. They are inverse permutations, so reading one as the other returns a different winner:

```
intent: C2 > C0 > C1
  pick-slot ballot [2,0,1]  read as ranked    -> C0 > C2 > C1   WRONG
  ranked ballot    [1,0,2]  read as pick-slot -> C1 > C0 > C2   WRONG
```

Since `type` is empty for a raw protocol, there is no signal to disambiguate, and no amount of inference recovers it — the information was never sent. `@vocdoni/ballot` consequently cannot decode a ranked result at all: it labels the question `multichoice` and reports "how many voters ranked each option", which is the same number for every option. Tracked downstream as vocdoni/integrator-sdk#22.

`saas-integrator-demo` works around this by storing its own `kind: 'ranked'` in its database — i.e. every integrator has to keep a private sidecar of the thing the API declined to record.

## How we got here

Worth stating, since it explains why this is worth closing rather than leaving as a TODO.

Because ranked has no named type, it is expressed as a raw `ballotProtocol` with `uniqueValues: true`. That put `uniqueValues` in front of people as "the flag you set for ranked-ish ballots" — and `QuestionTypeSetup` has a `UniqueChoices` field sitting right next to `MaxChoices`, which reads like the multichoice equivalent of the same idea. It is not: on the dense layout `multichoice` derives, uniqueness over 0/1 fields is unsatisfiable above two choices.

That is exactly #619 — elections that accepted votes and tallied zero. It is fixed now ([`api/processes.go#L189`](https://github.com/vocdoni/saas-backend/blob/9e811d07c7917e49ba10d39bdf477b816ef7fb78/api/processes.go#L189) rejects the flag, and the rejection message itself says *"use a ballotProtocol for a ranked ballot"*), but the shape of that fix is a signpost: the API is telling users to drop to a raw protocol because the named type they want does not exist. The `UniqueChoices` field remains on `QuestionTypeSetup` as a permanent trap that every named type must now explicitly reject.

Two silent all-zero-results bugs were found in the SDK in the same sweep (#619, plus budget/quadratic results being decoded with the wrong aggregation). Both were cases where the wire format looked determinate but the *meaning* was not, and both were invisible until a tally was decoded and compared against what voters actually picked. Named types are how that meaning gets recorded.

## Suggested scope

1. **`ranked`** — the clear one. Needs a `TypeSetup` that pins the convention the protocol cannot carry: at minimum whether the highest or lowest value wins (`saas-integrator-demo` uses highest, and integrator-sdk's docs now match it). Without that the axis stays ambiguous even with a type.
2. **`cumulative`** — budget / quadratic. `TypeSetup` would carry the total budget (`maxTotalCost`) and `costExponent`.
3. **`approval`** — decide whether it is real or a duplicate of `multichoice`. If it is a duplicate, drop the flag rather than implementing it, so the plan model stops advertising something with no meaning.
4. Once each exists, wire it into `OrgAllowsVotingType` so the plan flags gate for real, and reconsider whether `QuestionTypeSetup.UniqueChoices` should exist at all — if no named type ever accepts it, it is a field whose only behaviour is to be rejected.

Happy to do the integrator-sdk side (decoding, inference, React controls) once the type contract is decided here — that side is blocked on this, not the other way round.

lucasmenendez:
Closes #632.

## Problem

The plan model advertises six voting types, but the question API implemented only two
(`singlechoice`, `multichoice`). `ranked` and `cumulative`/quadratic were reachable only as a raw
`ballotProtocol`, with two consequences:

1. **Plan entitlements did not gate.** `OrgAllowsVotingType` mapped only the named types, so an org
   whose plan lacked `Ranked`/`Cumulative` could mint one anyway by sending a raw protocol. The flags
   sold nothing.
2. **The ballot type was not recoverable downstream.** A raw protocol carries no type label, so the
   integrator-sdk cannot decode it correctly, and every integrator keeps a private sidecar of what the
   API declined to record.

## Change

Gives `ranked` and `cumulative` named types by extending the forward/reverse machinery from #620
(`BallotProtocolFromType` / `QuestionTypeFromBallotProtocol`), then wires them into the plan gate.
Recognition stays "exact equality against the forward table," so the gate closes automatically once
the reverse function knows the shapes.

The four named types now have disjoint protocol images (pinned by `TestBallotShapeUnambiguous`):

| type | protocol image (N = choices) |
|---|---|
| `singlechoice` | `{maxCount:1, maxValue:max(value)}` |
| `multichoice` | `{maxCount:N, maxValue:1, costExponent:1, maxTotalCost:maxChoices}` |
| `ranked` | `{maxCount:N, maxValue:N-1, uniqueValues:true}` |
| `cumulative` | `{maxCount:N, maxValue:0, costExponent:1|2, maxTotalCost:budget}` |

- **ranked** — `maxValue 0` is unused; uniqueValues distinguishes it. Direction (highest-value-wins)
  is a global convention matching integrator-sdk docs, not encoded.
- **cumulative** — covers budget (`costExponent:1`) and quadratic (`costExponent:2`) via new
  `typeSetup.budget` / `typeSetup.costExponent` fields; one `Cumulative` plan flag.
- **approval** — intentionally no type: it is byte-identical to the multichoice dense layout, so the
  plan flag is documented as subsumed by `Multiple` rather than gating anything itself.
- **uniqueChoices** — kept (still rejected for multichoice); its error now points at `type: ranked`.

## Breaking changes

- **`saas-integrator-demo`'s current request is rejected.** It sends `type:"singlechoice"` + raw
  `{maxCount:3,maxValue:2,uniqueValues:true}`; the protocol now resolves to `ranked`, contradicting
  the `singlechoice` label → 400. Fix: send `type:"ranked"` (or omit `type`).
- A raw ranked/cumulative protocol now reads back with a **named `type`** (was `""`) and is **gated**
  by `plan.VotingTypes.Ranked`/`Cumulative`. No migration (same stance as #620).

## Verification

`go build`/`go vet` clean; `golangci-lint` 0 issues; `make swagger` regenerated. Unit tests in
`account/` (forward/reverse/`ResolveBallotShape`/`EffectiveQuestionType`/`TestBallotShapeUnambiguous`)
and `subscriptions/` pass; full `./api/` + `./db/` suites pass.

## Deploy note

`plan.VotingTypes` is parsed from Stripe product metadata (`stripe/service.go`); products whose
`votingTypes` JSON omits `ranked`/`cumulative` unmarshal them to `false`. Those flags gated nothing
before this PR, so live metadata may not set them. Already-minted questions are grandfathered (the
publish voting-type gate skips questions with an `upstreamId`), but for **new** questions the
relevant Stripe product metadata must be audited/updated to grant `ranked`/`cumulative` where
intended.

## Follow-up (deferred)

The voting-type gate runs only at `/processes` publish preflight. `POST /transactions` and the
legacy `/process` build path route through `HasTxPermission`, which does not (yet) check voting type
— so an org whose plan lacks a flag can still get that ballot signed through those doors. Feasible to
close (the tx carries the full ballot shape); tracked as a `TODO:future` in `hasElectionMetadataPermissions`.

### DIFF
--- a/account/ballot.go
+++ b/account/ballot.go
@@ -207,6 +207,18 @@ func ValidateBallotProtocol(bp *db.BallotProtocol) error {
 			uint64(bp.MaxValue)+1, bp.MaxCount, bp.MaxCount, bp.MaxValue,
 		)
 	}
+	// maxValue 0 is the protocol's "values are aggregable amounts" marker: the per-value cap then
+	// comes from maxTotalCost (raised to costExponent) or, for a weighted election, the census
+	// weight (costFromWeight). With neither set the chain imposes no limit at all — a voter can put
+	// any uint32 on any field — so refuse it. Scoped to MaxCount > 1: at one field, maxValue 0 is
+	// also what a singlechoice over a single 0-valued choice derives, which is a legitimate ballot
+	// rather than the amounts-mode evasion this catches.
+	if bp.MaxValue == 0 && bp.MaxTotalCost == 0 && !bp.CostFromWeight && bp.MaxCount > 1 {
+		return fmt.Errorf(
+			"maxValue 0 with maxTotalCost 0 and costFromWeight false leaves every field unbounded: " +
+				"a cumulative ballot needs a budget (maxTotalCost > 0) or costFromWeight",
+		)
+	}
 	return nil
 }
 
--- a/account/ballot.go
+++ b/account/ballot.go
@@ -144,9 +144,12 @@ func QuestionTypeFromBallotProtocol(
 	if bp == nil || len(choices) == 0 {
 		return "", db.QuestionTypeSetup{}, false
 	}
-	// At most one candidate can ever match, whatever the order: each named type's image is disjoint
-	// from the others — ranked alone sets uniqueValues, cumulative alone sets maxValue 0,
-	// singlechoice alone sets maxCount 1 — pinned by TestBallotShapeUnambiguous.
+	// At most one candidate can ever match, whatever the order: each named type maps to a protocol
+	// image no other named type produces, so recognition is a function rather than a first-match
+	// heuristic. No single field is unique to one type — cumulative shares maxCount 1 with
+	// singlechoice at one choice, and maxValue 0 with singlechoice when its only choice is valued 0
+	// — but the full 7-field image is distinct for every pair (costExponent and maxTotalCost settle
+	// those degenerate cases), pinned by TestBallotShapeUnambiguous.
 	candidates := []struct {
 		qType string
 		setup db.QuestionTypeSetup
--- a/account/ballot.go
+++ b/account/ballot.go
@@ -6,8 +6,9 @@ import (
 	"github.com/vocdoni/saas-backend/db"
 )
 
-// A question describes its ballot twice: as a named type (singlechoice/multichoice) with its
-// TypeSetup, and as a raw BallotProtocol. This file keeps the two halves in step.
+// A question describes its ballot twice: as a named type (singlechoice, multichoice, ranked,
+// cumulative) with its TypeSetup, and as a raw BallotProtocol. This file keeps the two halves
+// in step.
 //
 // BallotProtocolFromType is the single mapping table, and the reverse direction is defined as
 // equality against it — QuestionTypeFromBallotProtocol asks "which named type would produce
@@ -54,14 +55,20 @@ func VoteTypeFromQuestion(q *db.VotingProcessQuestion) (db.VoteType, error) {
 }
 
 // BallotProtocolFromType derives the canonical on-chain ballot parameters of a named question
-// type. singlechoice picks one of N choices (one field, value = the chosen Choice.Value);
-// multichoice is approval-style (one field per choice, each 0/1, with MaxTotalCost bounding the
-// number of selections).
+// type:
+//   - singlechoice picks one of N choices (one field, value = the chosen Choice.Value);
+//   - multichoice is approval-style (one field per choice, each 0/1, with MaxTotalCost bounding
+//     the number of selections);
+//   - ranked assigns every choice a distinct rank (one field per choice holding its rank 0..N-1,
+//     so uniqueValues holds); the convention is highest-value-wins;
+//   - cumulative spreads a budget across the choices (maxValue 0 is the "amounts" marker; the
+//     per-value cap comes from MaxTotalCost and CostExponent, 1 for a linear budget or 2 for
+//     quadratic).
 //
 // setup.MinChoices has no protocol counterpart and is ignored — the chain has no minimum-count
-// field. setup.UniqueChoices is ignored too: a named type never produces a unique-values ballot,
-// because on the multichoice layout it would admit no vote (#619) and on singlechoice it is
-// vacuous. Callers reject it at the API boundary rather than silently dropping it.
+// field. setup.UniqueChoices is ignored too: the only named type that uses uniqueValues is ranked,
+// and it derives the flag from the type, not from this field. Callers reject a multichoice that
+// sets it at the API boundary rather than silently dropping it.
 //
 // MaxVoteOverwrites and CostFromWeight are always zero here: QuestionTypeSetup has no field that
 // could express them, so a protocol carrying either is simply not a named shape.
@@ -89,6 +96,32 @@ func BallotProtocolFromType(
 			CostExponent: 1,
 			MaxTotalCost: setup.MaxChoices,
 		}, nil
+	case db.VotingTypeRanked:
+		// Ranking needs at least two choices to mean anything. maxValue reaches N-1 so the N
+		// distinct ranks (0..N-1) fit, which is also what makes uniqueValues satisfiable.
+		if len(choices) < 2 {
+			return nil, fmt.Errorf("ranked ballot requires at least two choices")
+		}
+		return &db.BallotProtocol{
+			MaxCount:     uint32(len(choices)),
+			MaxValue:     uint32(len(choices) - 1),
+			UniqueValues: true,
+		}, nil
+	case db.VotingTypeCumulative:
+		// maxValue 0 is the protocol's "values are aggregable amounts" marker: each field holds a
+		// credit, bounded in total by MaxTotalCost (the budget) raised to CostExponent. CostExponent
+		// 1 is a linear budget, 2 is quadratic.
+		if setup.Budget == 0 {
+			return nil, fmt.Errorf("cumulative ballot requires a budget greater than zero")
+		}
+		if setup.CostExponent != 1 && setup.CostExponent != 2 {
+			return nil, fmt.Errorf("cumulative ballot costExponent must be 1 (linear) or 2 (quadratic)")
+		}
+		return &db.BallotProtocol{
+			MaxCount:     uint32(len(choices)),
+			CostExponent: setup.CostExponent,
+			MaxTotalCost: setup.Budget,
+		}, nil
 	default:
 		return nil, fmt.Errorf("unsupported question type %q", qType)
 	}
@@ -99,8 +132,9 @@ func BallotProtocolFromType(
 // recognised only when BallotProtocolFromType reproduces the protocol exactly, so a recognised
 // type is guaranteed to re-derive the same on-chain parameters.
 //
-// ok is false for the shapes that have no named type — ranked, quadratic, anything using vote
-// overwrites or weighted cost — which stay expressible as a raw protocol.
+// ok is false for the shapes that have no named type — anything using vote overwrites or weighted
+// cost, or a cumulative/ranked protocol a client hand-crafted away from its canonical shape —
+// which stay expressible as a raw protocol.
 //
 // The choices are part of the question: a protocol whose MaxValue cannot carry the highest
 // Choice.Value is not singlechoice over those choices, however much it looks like one.
@@ -110,8 +144,9 @@ func QuestionTypeFromBallotProtocol(
 	if bp == nil || len(choices) == 0 {
 		return "", db.QuestionTypeSetup{}, false
 	}
-	// At most one candidate can ever match, whatever the order: CostExponent is unconditionally
-	// 0 for singlechoice and 1 for multichoice, so their images are always distinct.
+	// At most one candidate can ever match, whatever the order: each named type's image is disjoint
+	// from the others — ranked alone sets uniqueValues, cumulative alone sets maxValue 0,
+	// singlechoice alone sets maxCount 1 — pinned by TestBallotShapeUnambiguous.
 	candidates := []struct {
 		qType string
 		setup db.QuestionTypeSetup
@@ -120,6 +155,11 @@ func QuestionTypeFromBallotProtocol(
 		{db.VotingTypeMultiChoice, db.QuestionTypeSetup{
 			MinChoices: minChoicesFor(bp.MaxTotalCost),
 			MaxChoices: bp.MaxTotalCost,
+		}},
+		{db.VotingTypeRanked, db.QuestionTypeSetup{}},
+		{db.VotingTypeCumulative, db.QuestionTypeSetup{
+			Budget:       bp.MaxTotalCost,
+			CostExponent: bp.CostExponent,
 		}},
 	}
 	for _, candidate := range candidates {
@@ -213,9 +253,9 @@ type BallotShape struct {
 // half alone, and what keeps the API's answer equal to what the election will be.
 //
 // Supplying both halves is only allowed when they agree, or when the protocol has no named shape
-// (the ranked ballot a client may still label singlechoice). Two named shapes that disagree are an
-// error rather than a silent win for the protocol: a client editing a question through its
-// TypeSetup would otherwise get a 200 and none of its edit.
+// at all (vote overwrites, weighted cost, or a hand-crafted non-canonical shape). Two named shapes
+// that disagree are an error rather than a silent win for the protocol: a client editing a
+// question through its TypeSetup would otherwise get a 200 and none of its edit.
 //
 // MinChoices is the one thing that cannot be derived, having no on-chain counterpart, and is
 // carried over from the input for multichoice only, clamped to the resolved MaxChoices. A
--- a/account/ballot_test.go
+++ b/account/ballot_test.go
@@ -288,8 +288,9 @@ func TestBallotShapeUnambiguous(t *testing.T) {
 				protos = append(protos, named{qType, *bp})
 			}
 			// every named type well-defined over cs. A single multichoice representative suffices:
-			// maxChoices only changes its MaxTotalCost, and multichoice is the only type with
-			// maxValue 1, so no value of it can collide with another type's image.
+			// maxChoices only changes its MaxTotalCost, and multichoice is the only type whose image
+			// always carries uniqueValues:false with costExponent:1, so it can never collide with
+			// ranked (uniqueValues:true) or singlechoice (costExponent:0) for any maxTotalCost.
 			add(db.VotingTypeSingleChoice, db.QuestionTypeSetup{MinChoices: 1, MaxChoices: 1})
 			add(db.VotingTypeMultiChoice, db.QuestionTypeSetup{MaxChoices: uint32(n)})
 			if n >= 2 { // ranked needs at least two choices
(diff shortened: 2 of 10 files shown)


account/ballot.go
```diff
@@ -110,8 +144,9 @@ func QuestionTypeFromBallotProtocol(
 	if bp == nil || len(choices) == 0 {
 		return "", db.QuestionTypeSetup{}, false
 	}
-	// At most one candidate can ever match, whatever the order: CostExponent is unconditionally
-	// 0 for singlechoice and 1 for multichoice, so their images are always distinct.
+	// At most one candidate can ever match, whatever the order: each named type's image is disjoint
+	// from the others — ranked alone sets uniqueValues, cumulative alone sets maxValue 0,
+	// singlechoice alone sets maxCount 1 — pinned by TestBallotShapeUnambiguous.
```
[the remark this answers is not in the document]
lucasmenendez: Good catch — the rationale was wrong. No single field is unique to one type (cumulative shares `maxCount:1` with singlechoice at one choice, and `maxValue:0` with singlechoice when its only choice is valued 0); the images are distinct only as a full 7-field tuple (`costExponent`/`maxTotalCost` settle those degenerate cases), which `TestBallotShapeUnambiguous` pins for counts 1–8. Reworded in acb0f53.

subscriptions/subscriptions.go
```diff
@@ -110,7 +110,8 @@ func hasElectionMetadataPermissions(process *models.NewProcessTx, plan *db.Plan)
 		return false, fmt.Errorf("duration is greater than the allowed")
 	}
 
-	// TODO:future check if the election voting type is supported by the plan
+	// The election voting type is gated per-question at publish preflight (OrgAllowsVotingType),
```
elboletaire: **The TODO this replaces was accurate; the replacement claim is not.**

`hasElectionMetadataPermissions` is reached from `HasTxPermission`, which is called by `api/transaction.go:100` — the generic remote-signer endpoint where an org admin POSTs an arbitrary `NewProcessTx`. Publish preflight (`publishPreflightProblems` → `OrgAllowsVotingType`) never runs on that path, so `plan.VotingTypes.{Single,Multiple,Ranked,Cumulative}` are still enforced *nowhere* for it.

Concrete scenario: an org whose plan has `ranked: false` builds a NewProcessTx with `VoteOptions{maxCount:3, maxValue:2}` + `EnvelopeType{uniqueValues:true}` and POSTs it to `/transactions`. It is signed, funded and mined — the exact evasion this PR set out to close, just through the other door.

The stated reason is also at odds with the PR's own thesis: the tx *does* carry the ballot shape (`VoteOptions` + `EnvelopeType`), which is precisely what `QuestionTypeFromBallotProtocol`/`EffectiveQuestionType` recognise. `OrgCanPublishProcess`'s godoc still says "the authoritative enforcement remains HasTxPermission at build time", which is now false for voting type. Please keep a TODO here (or map the tx's VoteOptions back through the recogniser) rather than asserting the gate lives elsewhere.

api/processes_publish.go
```diff
@@ -109,19 +109,16 @@ func (a *API) publishPreflightProblems(
...
-	// (maxValue == 1 && maxCount > 1 ⇒ effectively multiple-choice, whatever else is set), not
-	// this one.
+	// What still slips through is a protocol that is *almost* canonical: {maxCount:2, maxValue:1,
+	// maxTotalCost:2} is the multichoice ballot field-for-field bar costExponent, so it is not
+	// recognised and stays ungated. EffectiveQuestionType recognises a shape only when it is
+	// exactly canonical, which is what storage needs and what authorization does not. Actually
+	// holding the gate needs a deliberately loose classifier (maxValue == 1 && maxCount > 1 ⇒
+	// effectively multiple-choice, whatever else is set), not this one.
 	for i := range questions {
 		if err := a.subscriptions.OrgAllowsVotingType(vp.OrgAddress, account.EffectiveQuestionType(&questions[i])); err != nil {
```
elboletaire, line 129: **A partially-published process can become permanently unpublishable after this deploy.**

The publish handler explicitly supports resuming: `if !anyMined(questions)` at `processes_publish.go:277` exists so "a re-publish of a process that already mined some elections" works. But this preflight runs *before* the claim, and it now denies the whole process when any question's effective type lacks its plan flag.

Scenario: on `main`, an org whose plan has `ranked: false` publishes a draft carrying the raw permutation protocol (`{maxCount:3, maxValue:2, uniqueValues:true}` — the shape `saas-integrator-demo` sends). Question 1 mints on chain, question 2 fails to fund, `vp.Published` stays false. This PR deploys. Every subsequent publish attempt now returns 400 `"voting type not allowed (ranked)"`, so the remaining questions can never be minted while question 1's election is live on chain.

Compounding it: `plan.VotingTypes` is parsed from Stripe product metadata (`stripe/service.go:165`, `extractPlanMetadata[db.VotingTypes]`), so any product whose `votingTypes` JSON simply omits `ranked`/`cumulative` unmarshals them to `false`. As the PR description itself notes, those flags "sold nothing" until now, so there is no reason to expect live metadata to have them set — the repo's own `mockEssentialPlan` has `Ranked: false`. Worth auditing the live Stripe metadata (or grandfathering questions that already have an `UpstreamID`) before merging.

account/ballot.go
```diff
@@ -89,6 +96,32 @@ func BallotProtocolFromType(
...
+		return &db.BallotProtocol{
+			MaxCount:     uint32(len(choices)),
+			MaxValue:     uint32(len(choices) - 1),
+			UniqueValues: true,
+		}, nil
+	case db.VotingTypeCumulative:
+		// maxValue 0 is the protocol's "values are aggregable amounts" marker: each field holds a
+		// credit, bounded in total by MaxTotalCost (the budget) raised to CostExponent. CostExponent
+		// 1 is a linear budget, 2 is quadratic.
+		if setup.Budget == 0 {
```
elboletaire, line 114: **Both new cumulative rules are trivially bypassed by a raw protocol, which also keeps the ballot ungated.**

`budget > 0` is enforced only on the named path. Send instead:

```json
{"ballotProtocol": {"maxCount": 3, "maxValue": 0, "costExponent": 1, "maxTotalCost": 0}}
```

- `ValidateBallotProtocol` passes it (`maxCount != 0`, no `uniqueValues`).
- `QuestionTypeFromBallotProtocol` does *not* recognise it: the cumulative candidate uses `Budget: bp.MaxTotalCost = 0`, so `BallotProtocolFromType` errors and the candidate is skipped. Stored `Type` is `""`.
- `EffectiveQuestionType` returns `""` → `OrgAllowsVotingType` short-circuits to `nil`, so an org whose plan lacks `Cumulative` mints it anyway.
- On chain this is the worst variant: `maxValue == 0` puts `results.go` into aggregate-amounts mode, and `maxTotalCost == 0` with `costFromWeight == false` skips the cost check entirely (`if r.VoteOpts.MaxTotalCost > 0 || r.EnvelopeType.CostFromWeight`). A voter can put any `uint32` on any field.

So the one shape the named path refuses as meaningless is exactly the one the raw path accepts *and* leaves unmetered. If `budget == 0` is genuinely not a countable ballot, `ValidateBallotProtocol` is the place to say so (`maxValue == 0 && maxTotalCost == 0 && !costFromWeight` ⇒ no bound on any value).

account/ballot.go
```diff
@@ -121,6 +159,11 @@ func QuestionTypeFromBallotProtocol(
 			MinChoices: minChoicesFor(bp.MaxTotalCost),
 			MaxChoices: bp.MaxTotalCost,
 		}},
+		{db.VotingTypeRanked, db.QuestionTypeSetup{}},
+		{db.VotingTypeCumulative, db.QuestionTypeSetup{
```
elboletaire, line 163: **`minChoices`/`maxChoices` sent with a cumulative question are silently dropped, with no error.**

The cumulative candidate setup carries only `Budget`/`CostExponent`, and the `MinChoices` carry-over at line 301 is gated on `qType == db.VotingTypeMultiChoice`. Combined with the new `case db.VotingTypeSingleChoice, db.VotingTypeRanked, db.VotingTypeCumulative:` no-op in `buildQuestions`, nothing validates or preserves them.

Scenario: a client POSTs `{"type":"cumulative","typeSetup":{"minChoices":2,"maxChoices":3,"budget":12,"costExponent":2}}`. It gets a 200, and the stored/returned `typeSetup` is `{"minChoices":0,"maxChoices":0,"budget":12,"costExponent":2}` — the client believes it constrained the spread across choices and it did not. Multichoice goes out of its way to 400 on `minChoices > maxChoices` and to preserve `MinChoices`; cumulative accepts the same fields and discards them.

Either reject a cumulative `typeSetup` that sets `minChoices`/`maxChoices` (the way multichoice rejects `uniqueChoices`), or carry `MinChoices` through as multichoice does. Same applies to `ranked`, whose resolved setup is `{}`, so `uniqueChoices: true` sent with it is also accepted-and-dropped while multichoice 400s on it.

account/ballot_test.go
```diff
@@ -227,23 +269,38 @@ func TestQuestionTypeFromBallotProtocol(t *testing.T) {
...
+			var protos []named
+			add := func(qType string, setup db.QuestionTypeSetup) {
+				bp, err := BallotProtocolFromType(qType, setup, cs)
 				c.Assert(err, qt.IsNil)
-				c.Assert(*single, qt.Not(qt.Equals), *multi,
-					qt.Commentf("n=%d maxChoices=%d", n, maxChoices))
+				protos = append(protos, named{qType, *bp})
+			}
+			// every named type well-defined over cs. A single multichoice representative suffices:
+			// maxChoices only changes its MaxTotalCost, and multichoice is the only type with
```
elboletaire: The justification for reducing multichoice to a single representative is factually wrong: `ranked` over 2 choices derives `{maxCount:2, maxValue:1, uniqueValues:true}`, so multichoice is *not* "the only type with maxValue 1" — and `n=2` is inside this loop's range.

The reduction still holds, but for a different reason: multichoice always carries `uniqueValues:false` + `costExponent:1`, which separates it from ranked (`uniqueValues:true`) and from singlechoice (`costExponent:0`) for *every* `maxTotalCost`. Since this comment is what licenses dropping the old `for maxChoices := range uint32(n+1)` sweep, leaving it stated incorrectly invites a future edit that reduces the test unsoundly on a premise that was never true.

subscriptions/subscriptions.go
lucasmenendez: Deferred per scope, with the gap now stated honestly rather than claimed closed. Restored a `TODO:future` in `hasElectionMetadataPermissions` noting the voting-type gate runs only at `/processes` publish preflight, so `POST /transactions` and the legacy `/process` build path (both → `HasTxPermission`) bypass it; and corrected the now-false line in `OrgCanPublishProcess`'s godoc ("authoritative enforcement remains HasTxPermission" was true for census/count/weighted/duration, not voting type). Feasibility confirmed — all 7 BallotProtocol fields are on the tx and choices aren't a blocker — so this is a deliberate follow-up, not blocked. (7a33d81)

api/processes_publish.go
lucasmenendez: Fixed: `publishPreflightProblems` now skips the voting-type gate for questions with a non-empty `UpstreamID` — the same predicate the codebase already uses as the immutability boundary (`reconcileDraftQuestionShapes`, the mint loop, `ResetQuestionsPublish`). It lives in the shared function, so publish and the validation dry-run agree, and a resume is no longer wedged by a plan change after a partial publish. On the Stripe point: grandfathering covers already-minted questions; for new questions, product metadata that omits ranked/cumulative still unmarshals to false — flagged as a deploy-audit step in the PR body. Test added (`TestVotingProcessPublishGateSkipsMinted`). (7a33d81)

account/ballot.go
lucasmenendez: Fixed in `ValidateBallotProtocol`: it now rejects `maxValue 0 && maxTotalCost 0 && !costFromWeight` — the unmetered amounts shape. One refinement from your literal rule: scoped to `MaxCount > 1`, because at one field `maxValue 0` is also what a singlechoice over a single 0-valued choice derives (`BallotProtocolFromType` sets `MaxValue = max(Choice.Value)`), and rejecting it would make a one-choice singlechoice unpublishable. With `MaxCount > 1` the only sensible reading is amounts mode, so the evasion is caught without that regression. Test added ("unbounded cumulative"). (7a33d81)

account/ballot.go
lucasmenendez: Fixed: `buildQuestions` now rejects spurious `typeSetup` rather than dropping it silently — ranked rejects any non-empty setup (it uses none), cumulative rejects `minChoices`/`maxChoices`/`uniqueChoices` (it uses only `budget`/`costExponent`), mirroring how multichoice already rejects `uniqueChoices`. singlechoice's documented "ignores typeSetup" is left as-is (changing it is out of scope). Tests added. (7a33d81)

account/ballot_test.go
lucasmenendez: Fixed: reworded to the real reason — multichoice is the only type whose image always carries `uniqueValues:false` with `costExponent:1`, so it can never collide with ranked (`uniqueValues:true`) or singlechoice (`costExponent:0`) for any `maxTotalCost`. Not "the only type with maxValue 1". (7a33d81)

elboletaire (approved):

Approving. `7a33d81` addresses the review; re-checked against that commit.

**Fixed**

- `account/ballot.go` — the unmetered-amounts hole is closed. `{maxCount:3, maxValue:0, costExponent:1, maxTotalCost:0}` is now rejected, and the `maxCount > 1` scoping keeps the legitimate singlechoice-over-one-0-valued-choice shape passing. `costFromWeight:true` still passing is correct: dvote runs the cost check when `MaxTotalCost > 0 || CostFromWeight`, so the census weight bounds it.
- `api/processes.go` — spurious `typeSetup` is a 400 now instead of being silently dropped.
- `account/ballot_test.go` — the disjointness comment cites the actual separator.
- `subscriptions/subscriptions.go` — the TODO is honest again and `OrgCanPublishProcess`'s godoc no longer claims `HasTxPermission` gates voting type. `POST /transactions` and the legacy `/process` path stay ungated, which is a deliberate deferral rather than a fix — fine to ship that way as long as it's tracked.

**One residual, not blocking**

`api/processes_publish.go` skips mined questions, which handles the already-minted case. The gate still runs on every *pending* question, so a draft whose remaining questions carry the newly-disallowed type still 400s on resume with a live on-chain election and `vp.Published` false. Since `plan.VotingTypes` comes from Stripe metadata where an omitted key unmarshals to `false`, that's the likelier direction of drift. The consistent reading of the commit's own reasoning is to skip the gate whenever `anyMined(questions)` — the process is committed on chain either way. Worth a follow-up, not a blocker.

